PR #2, opened 2019-11-19
Testing with JTAG-HS3

Martoni:
Works


trabucayre:
Applied, thanks.
Gwen

PR #10, opened 2019-12-29
fix gowin configuration command in readme

Martoni:
Not tested but I think it was wrong.


trabucayre:
True!
Applied! Thanks

PR #16, opened 2020-01-23
Update Makefile

Martoni:
No C files in project.


Martoni:
useless since CMake used

PR #18, opened 2020-02-16
Print error when no bitfile is specified

japm48:
This includes the case when no argument is given, but `--list-boards`, etc. should still work.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -49,6 +49,7 @@ struct arguments {
 	bool list_fpga;
 	bool write_flash;
 	bool write_sram;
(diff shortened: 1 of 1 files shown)


japm48:
Ok, it should be fixed now, hopefully.
Basically what bothered me is that executing without arguments resulted in a programming attempt.

trabucayre:
Now It's okay for me.
It's true, program() method is always called, but if no bitstream is provided, this method return immediatly. I will change this to call program() only when bit_file is not empty. I think it's less error prone.
Applied, thanks.

PR #19, opened 2020-02-16
make installation directory configurable

japm48:
This removes the hardcoded paths to allow configurability (useful for packaging).

With this patch, the paths are passed as macros added at compiler invocation; in particular, the macro `DATA_DIR` is, by default, `"/usr/local/share"`. This is done by taking advantage of the GNUInstallDirs feature of CMake to locate the `.../share` directory.

In this case, this should cleaner than a generated file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@
 cycloader
 openFPGALoader
 *.swp
+
+# Build directory
+build/
+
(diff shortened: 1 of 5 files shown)


trabucayre:
Thanks a lot for this PR.
Applied.

PR #20, opened 2020-02-16
Add support for Arty A7-100T

japm48:
Added new part: xc7a100.

This works directly when using `--board=arty`.


trabucayre:
Applied. Thanks

trabucayre:
I must generates spiOverJtag bit_file for this mode now :)

PR #29, opened 2020-04-16
Fixed DangerousPrototype's Bus Blaster support (in JTAGkey mode).

udif:
The Bus Blaster uses an FT2232H with an on-board CPLD allowing it to
impersonate various JTAG probes.
This did not work until now because JTAGkey uses ADBUS4/GPIOL0 as ~CS
to enable the JTAG buffers on the on-board CPLD.


trabucayre:
Thank for your contribution. But,
- there are 4 version of the bus blaster, these additions are true for all of them?
- can you add this two entries in alphabetical order?
Thank again

udif:
These are the revisions:
- V1 is using a hardware buffer for the JTAG signals, emulating an [Amontec  JTAGkey](https://web.archive.org/web/20140210101642/http://www.amontec.com/jtagkey.shtml).
- V2 is using an XC2C32A CPLD to enable emulating various FT2232H based products.
- V3 is a minor revision (form factor change), FT2232H clock output added as an CPLD input
- V4 Moved to a larger CPLD to enable SWV mode for ARM debugging

I have v2.0a1, but from all I read, the change should work on all of them.
All revisions requires the ADBUS4/GPIOL0 output from the FFT2232H to be active low to enable the JTAG outputs, even if the implementation is different (hardwired logic/different CPLDs).

I'll change the entries to be in alphabetical orders.

trabucayre:
Applied. Thanks

PR #33, opened 2020-05-12
Add support for ECPIX-5

jeanthom:
What was tested: flashing a bitstream in SRAM (OK) and onto the SPI flash (OK)


trabucayre:
Looks good for me. But could you update the README to add this board and cable?

Thanks

jeanthom:
I added the board to the README, but I'm not sure if I should include it as a cable too (the debug interface is integrated to the board)?

trabucayre:
Applied
Thanks

No it's true this is an internal programmer so not mandatory to at details.

PR #50, opened 2020-09-22
Add FTDI channel selection command line argument.

ronyrus:
Some boards have FTDI JTAG connected to the FPGA on different channel.
It's helpful to be able to provide the channel on command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ openFPGALoader -- a program to flash cyclone10 LP FPGA
   -b, --board=BOARD          board name, may be used instead of cable
   -c, --cable=CABLE          jtag interface
   -d, --device=DEVICE        device to use (/dev/ttyUSBx)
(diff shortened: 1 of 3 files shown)


trabucayre:
True. It's a great idea.
I've just two comment:

- part added in the main function and the function `ftdi_update_channel` are primarly sanity check, I prefer this part in part_opt (as done for [pins](https://github.com/trabucayre/openFPGALoader/blob/master/src/main.cpp#L320)). The simpliest way is to check if ID are correct and update ftdi_channel with the ID of the channel. So, in the `main` it's just mandatory to check if this option is != -1 and if it's an ftdi cable before updating cable.config.interface (without new function addition).
- the README must be updated to reflect this new option

Thanks

ronyrus:
done
was looking at the readme and it seems that "-d" is no longer with us...

trabucayre:
Look Good.
Applied thanks

trabucayre:
-d or --device is used if udev support is enable.

PR #51, opened 2020-09-29
add ftdi serial number argument

martin2250:
Hi again,

I needed a way to work with multiple FTDI devices attached to a single PC, so I added an option to specify the serial number (USB string descriptor) for selecting the right device.

There are a few other changes related to the way the JTAG device is constructed. Previously, there were two constructors for everything, with and without the `dev` parameter. Since, ultimately, there are no differences between calling the constructor with an empty string as the dev parameter or calling the other constructor without the dev parameter, I removed the constructors without the dev parameter, as to not add even more code overhead.

The one actually relevant change is in 73 of `src/ftdipp_mpsse.cpp`.

Let me know your thoughts!

Martin

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,28 +108,34 @@ After that you need to unplug and replug your device.
 ```bash
 openFPGALoader --help
 Usage: openFPGALoader [OPTION...] BIT_FILE
-openFPGALoader -- a program to flash cyclone10 LP FPGA
+openFPGALoader -- a program to flash FPGA
 
-  -b, --board=BOARD          board name, may be used instead of cable
-  -c, --cable=CABLE          jtag interface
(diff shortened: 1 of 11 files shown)


trabucayre:
Look good for me. Just two small remarks:
- ftdiJtagBitbang.cpp line 44 is too long
- the README must be updated with this new option
Thanks

martin2250:
Great! I just addressed both remarks. The Readme has more than one updated line, since the switch to cxxopts changed the --help message. I'm assuming sooner or later the proper --help message would have been copied to the Readme anyways.

trabucayre:
Applied.
Thank to have fully updated help in README

martin2250:
sure thing, thank you too!

PR #52, opened 2020-09-29
fix MACHXO3LF family

martin2250:
yeah, this should be pretty self-evident
https://github.com/trabucayre/openFPGALoader/blob/7b048d2d823ae70fd153bea44593e92c19bf275e/src/part.hpp#L35


trabucayre:
Thank to point this typo!
Applied

PR #61, opened 2020-11-22
Add uaccess tag to udev rules to work on distros without plugdev group

tdaede:
This automatically grants the device to the currently logged in user,
without the need for a group.

Leave the old plugdev group for backwards compatibility.


trabucayre:
Applied
Thanks!

PR #67, opened 2020-12-12
Fix gcc 10.2 error messages.

phdussud:
Hello! 
I noticed that there were quite a few warnings produced during debug compilations and they feel legitimate to me. I fixed them, so we don't have to worry about them. 
Thanks!


trabucayre:
LGTM
Applied
Thanks for these fixes!

PR #70, opened 2020-12-24
limit the progressBar update rate to 5 per second. This speeds up loa…

phdussud:
Hello! 
I would like to submit this small change to the progress bar. It limit its update rate to 5 updates a second. This does speed up the loading of small bin files because the terminal takes time to redisplay the bar. 
Thanks a lot. 
Patrick.

### DIFF
--- a/src/progressBar.cpp
+++ b/src/progressBar.cpp
@@ -17,6 +17,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include "time.h"
 #include "progressBar.hpp"
 #include "display.hpp"
 
@@ -24,9 +25,15 @@ ProgressBar::ProgressBar(std::string mess, int maxValue, int progressLen):
 		_mess(mess), _maxValue(maxValue), _progressLen(progressLen)
 {
 }
-
-void ProgressBar::display(int value)
+static time_t last_time; 
+void ProgressBar::display(int value, char force)
 {
+	clock_t this_time = clock();
+	if (!force && ((((float)(this_time - last_time))/CLOCKS_PER_SEC) < 0.2))
+	{
+		return;
+	}
+	last_time = this_time;
 	float percent = ((float)value * 100.0f)/(float)_maxValue;
 	float nbEq = (percent * (float) _progressLen)/100.0f;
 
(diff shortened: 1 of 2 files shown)


trabucayre:
Good idea!
Only one small remark: `last_time` must be a class member, with an initialization at constructor time, currently the first time `display` is called `last_time` has an unknown value.

Thanks

phdussud:
I agree! I will make the change. I just want to point out that right now the static has a well-defined value of 0 when the program starts. It is per C and C++ standard.
Thanks!
Patrick

phdussud:
Done! 
Thanks again, 
Patrick

trabucayre:
Sorry for delay.
LGTM
applied Thanks

phdussud:
Thank you very much. 
I have been working on adding a new cable, the breakout board for the Cypress FX2LP. A board costs about $8 at amazon 
https://www.amazon.com/EZ-USB-FX2LP-CY7C68013A-Analyzer-Source/dp/B081DH4TN3
It is about 2x slower than a good FT232H cable. These boards are popular as logic analyzers, as they are supported by sigrok.org
I have done the port using the excellent library libfpgalink : https://github.com/makestuff/libfpgalink
The downside of this port is that it adds external dependencies, namely the above git repo. I am not sure it is justified for a normal release of openFPGAloader that is very tight and relies only on distro supported dependencies. 
I see 3 ways around this:
- Drop this idea. 
- Make it a conditional compilation feature and leave it off by default. 
- Try to make libfpgalink delay loaded, which seems possible even under Linux using this component: https://github.com/yugr/Implib.so
The port itself is simple because your jtag abstraction matches fairly well with the libfpgalink API. 
Please let me know your thoughts on this. 
Patrick

trabucayre:
I'm open to all new cable (maybe I need to order one of this), and not oppossed to add dependencies to external libraries when it's possible instead of reimplementing everything from scratch (it's depends more or less to the complexity).
The idea of using an option to enable this cable seems good, these allows to have something simple for most people but allowed to use some features when it's needed.

phdussud:
Based on your comments, I went ahead and modified the cmake file to include FX2 support only when it can find libfpgalink on the path. This is analogous to the treatment of libudev.

trabucayre:
Maybe an `option` is required to let user decide if he want to add support to this cable or not. Of course, test about library is mandatory.
I'm not usually happy when a build system decide for me what to build according to libraries present in my OS.
I'm interested to see your modifications.

phdussud:
Sounds good. Do you prefer to have a look on my forked repo on github or for me to submit a PR?
Thanks,
Patrick

trabucayre:
The best solution is to update your fork and send a PR when you consider you are satisfied by your implementation.

PR #75, opened 2021-02-21
Add support for GW1NS-2C FPGA chips

Disasm:
I tested this with my custom board (https://github.com/Disasm/honeycomb-pcb) and the `-b tangnano` openFPGALoader option. Both SRAM and flash programming work. Should I add a separate board definition for my board? I don't think I will be able to manufacture these boards in the nearest future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,13 @@ __Current support kits:__
 * [Terasic de0nano](https://www.terasic.com.tw/cgi-bin/page/archive.pl?No=593) (memory)
 * LambdaConcept ECPIX-5 (memory and flash)
 * [Efinix Xyloni](https://www.efinixinc.com/products-devkits-xyloni.html) (spi flash (*xyloni_spi*))
+* [honeycomb](https://github.com/Disasm/honeycomb-pcb) (memory and internal flash)
 
 __Supported (tested) FPGA:__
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM.
Just a question is it not a `GW1N` variant ?
For the board as you want.
But why doing a clone using an STM32 instead of using something like @jeanthom 's [dirtyJTAG](https://github.com/jeanthom/DirtyJTAG) ? In fact I have many issues with the  tangNano ch552 (FT2232C clone)

Disasm:
It's not a GW1N variant, it's GW1NS-UX2CQN48C5/I4: GW1NS-2C family, 
I made an STM32F042 clone for multiple reasons:
* small size, cheap, easy to solder
* XTAL is not required for STM32F042, this lowers BOM cost even more. Clock for FPGA is generated by timer on STM32.

My implementation (https://github.com/Disasm/f042-ftdi) works better than that on TangNano. I've never succeed in programming flash on LonganNano, but with my FTDI emulation this works flawlessly.

trabucayre:
Interesting. Must test this!
Could you just add this device to the README ?

Thanks

trabucayre:
Applied thanks!

PR #76, opened 2021-02-24
Olimex iCE40HXXK-EVB - Pinout wrong in README?

gbuzogany:
I attempted to use openFPGALoader to program an iCE40HX8K-EVB using an FTDI FT232H. The flash/programming interface is almost identical to the iCE40HX1K-EVB, so it should work exactly the same way. I followed the wiring from the README, but it did not work:

```
|     FTDI     | iCE40HX1K-EVB (PGM1) |
|--------------|----------------------|
| SCK (ADBUS0) | Pin 9                |
| SI  (ADBUS1) | Pin 7                |
| SO  (ADBUS2) | Pin 8                |
| CS  (ABDUS4) | Pin 10               |
| RST (ADBUS6  | Pin 6                |
| DONE (ADBUS7)| Pin 5                |
```

I spent a few hours fiddling with it, and noticed that the Flash DO/DI is in swapped with the FPGA DO/DI (FPGA DO = Flash DI, vice-versa), so I attempted to invert pins 7 and 8, and then it worked perfectly.

Because of that... I guess the README is not right?
I updated the README to add the iCE40HX8K-EVB as a supported board (not sure if I should, but it works), and inverted the pins 7, 8 on the section specific to the iCE40HXXK-EVB boards to prevent other people from hitting the same problem.


trabucayre:
You have right, Thanks!
I have this board for a while, with a custom adapter, and when I've written this note I have used the schematic where pin name are from ice40 instead of SPI Flash... And done a mistake...

PR #77, opened 2021-02-26
Support for new DirtyJtag2 protocol

phdussud:
DirtyJTAG has a new more efficient protocol that allows for faster loads by reducing the USB traffic (now mostly one way writes instead of writes -> reads)

Thanks, 
Patrick


trabucayre:
Thanks!
Really interesting! I've not seen this branch before, must try it!
For purge buffer, is it possible to rebase to the HEAD to drop these commit ( or waiting to have a contrib ready to be merged to recreate this PR) ?
See further for my comments

src/dirtyJtag.cpp
```diff
@@ -172,31 +232,51 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
 		memset(tx_cpy, 0, real_byte_len);
 	tx_ptr = tx_cpy;
 
-	/* first send 30 x 8 bits */
-	tx_buf[0] = CMD_XFER;
+	tx_buf[0] = CMD_XFER | ((rx || (_version <= 1)) ? 0 : NO_READ);
+	uint16_t max_bit_transfer_length = (uint16_t[]){240, 496, 4000}[_version -1];
```
trabucayre, line 236: again using a struct with all configs to avoid to add complexity

src/dirtyJtag.cpp
```diff
@@ -172,31 +232,51 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
+		}
+		memset(tx_buf + header_offset, 0, byte_to_send);
 		for (int i = 0; i < bit_to_send; i++)
 			if (tx_ptr[i >> 3] & (1 << (i & 0x07)))
-				tx_buf[2 + (i >> 3)] |= (0x80 >> (i & 0x07));
-
-		tx_buf[2 + byte_to_send] = CMD_STOP;
-
+				tx_buf[header_offset + (i >> 3)] |= (0x80 >> (i & 0x07));
+
```
trabucayre, line 259: space in a blank line

src/dirtyJtag.cpp
```diff
@@ -172,31 +232,51 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
+				tx_buf[header_offset + (i >> 3)] |= (0x80 >> (i & 0x07));
+		
+		actual_length = 0;
 		int ret = libusb_bulk_transfer(dev_handle, DIRTYJTAG_WRITE_EP,
-		        (unsigned char *)tx_buf, 33, &actual_length, 1000);
-		if (ret < 0) {
-			cerr << "writeTDI: fill: usb bulk write failed " << ret << endl;
+		        (unsigned char *)tx_buf, (byte_to_send + header_offset), &actual_length, 1000);
+		if ((ret < 0) || (actual_length != (int)(byte_to_send + header_offset))) {
+			cerr << "writeTDI: fill: usb bulk write failed " << actual_length << endl;
```
trabucayre, line 264: when something wrong with `libusb_bulk_transfer` `ret` contains error code. Maybe useful to let ret in message

src/dirtyJtag.cpp
```diff
@@ -230,6 +310,7 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
 			}
 		}
 	}
+
```
trabucayre, line 313: why?

src/dirtyJtag.hpp
```diff
@@ -60,10 +60,13 @@ class DirtyJtag : public JtagInterface {
...
 
 	int sendBitBang(uint8_t mask, uint8_t val, uint8_t *read, bool last);
+	void getVersion();
 
     libusb_device_handle *dev_handle;
 	libusb_context *usb_ctx;
 	uint8_t _tdi;
 	uint8_t _tms;
+	uint8_t _version;
+
```
trabucayre, line 70: why?

trabucayre (changes requested):

src/dirtyJtag.cpp
```diff
@@ -23,17 +23,33 @@
...
+#ifdef DIRTY_JTAG_TEENSY_4
+
+#define DIRTYJTAG_VID 0x16C0 
+#define DIRTYJTAG_PID 0x1485 
+
+#define DIRTYJTAG_INTF		  0
+#define DIRTYJTAG_WRITE_EP    0x2 
+#define DIRTYJTAG_READ_EP     0x82
+
+#else
```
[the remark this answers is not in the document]
phdussud: It isn't officially supported, no.

src/dirtyJtag.cpp
```diff
@@ -71,17 +92,20 @@ DirtyJtag::DirtyJtag(uint32_t clkHZ, bool verbose):
...
 
-	if (clkHZ > 600000) {
-		printInfo("DirtyJTAG probe limited to 600kHz");
-		clkHZ = 600000;
+	_version = 0;
+	getVersion();
+
+	if (clkHZ > 16000000) {
+		printInfo("DirtyJTAG probe limited to 16MHz");
+		clkHZ = 16000000;
```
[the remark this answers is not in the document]
phdussud: It is compatible with dirtyJtag1, which has an internal maximum to the frequency, so it is ok to pass a higher frequency.

src/dirtyJtag.cpp
```diff
@@ -97,6 +121,42 @@ DirtyJtag::~DirtyJtag()
...
+	if (ret < 0) {
+		cerr << "getVersion: usb bulk write failed " << ret << endl;
+		return;
+	}
+	do {
+		ret = libusb_bulk_transfer(dev_handle, DIRTYJTAG_READ_EP,
+						rx_buf, 64, &actual_length, 1000);
+		if (ret < 0) {
+					cerr << "getVersion: read: usb bulk read failed " << ret << endl;
+					return;
```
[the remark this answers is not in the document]
phdussud: will fix

src/dirtyJtag.cpp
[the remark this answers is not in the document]
trabucayre: Maybe to avoid this it's possible to 
1. try to find official vid/pid;
2. if it fail trying teensy.
and using a variable to store write_ep value depending to the target

src/dirtyJtag.cpp
```diff
@@ -45,6 +61,11 @@ enum dirtyJtagCmd {
 	CMD_CLK =    0x06
 };
 
+enum CommandModifier {
+  EXTEND_LENGTH = 0x40,
+  NO_READ       = 0x80
+};
+
```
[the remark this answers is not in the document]
phdussud: Sure.  I chose to lift the code from DirtyJTAG2 so I wouldn't make a mistake but I can change it.

src/dirtyJtag.hpp
phdussud: will fix

src/dirtyJtag.cpp
phdussud: will fix

src/dirtyJtag.cpp
phdussud: sure. will fix

src/dirtyJtag.cpp
phdussud: will fix

src/dirtyJtag.cpp
phdussud: I am sorry but I don't follow....

src/dirtyJtag.cpp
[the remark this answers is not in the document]
phdussud: I took this code off. It isn't worth additional complexity. I wrote an implementation of Dirtyjtag for Teensy because that's the only processor I have that handles high speed USB. I wanted to know how much faster it would be compared to dirtyjtag2. I don't think teensy 4.x makes a particularly cost effective probe.

src/dirtyJtag.cpp
phdussud: I attempted to interpret your comment and code it..

src/dirtyJtag.cpp
```diff
@@ -224,8 +233,8 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
 		memset(tx_cpy, 0, real_byte_len);
 	tx_ptr = tx_cpy;
 
-	tx_buf[0] = CMD_XFER | ((rx || (_version <= 1)) ? 0 : NO_READ);
-	uint16_t max_bit_transfer_length = (uint16_t[]){240, 496, 4000}[_version -1];
+	tx_buf[0] = CMD_XFER | (rx ? 0 : v_options[_version].no_read );
+	uint16_t max_bit_transfer_length = v_options[_version].max_bits;
```
trabucayre, line 237: LGTM

src/dirtyJtag.cpp
```diff
@@ -93,14 +114,49 @@ DirtyJtag::~DirtyJtag()
...
 {
 	int actual_length;
 	int ret, req_freq = clkHZ;
 
-	if (clkHZ > 600000) {
-		printWarn("DirtyJTAG probe limited to 600kHz");
-		clkHZ = 600000;
+	if (clkHZ > 16000000) {
+		printWarn("DirtyJTAG probe limited to 16000kHz");
+		clkHZ = 16000000;
```
trabucayre, line 159: By using new struct it's maybe a good idea to fix max frequency using it

src/dirtyJtag.cpp
```diff
@@ -177,31 +233,51 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
+			tx_buf[0] &= ~EXTEND_LENGTH;
+			tx_buf[1] = bit_to_send;
+			header_offset = 2;
+		}
+		memset(tx_buf + header_offset, 0, byte_to_send);
 		for (int i = 0; i < bit_to_send; i++)
 			if (tx_ptr[i >> 3] & (1 << (i & 0x07)))
-				tx_buf[2 + (i >> 3)] |= (0x80 >> (i & 0x07));
-
-		tx_buf[2 + byte_to_send] = CMD_STOP;
```
trabucayre, line 191: CMD_STOP must be left. No ?

src/dirtyJtag.cpp
```diff
@@ -177,31 +233,51 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
+
+		if (rx || (_version <= 1)) {
+			int transfer_length = (bit_to_send > 255) ? byte_to_send :32;
+			do {
+				ret = libusb_bulk_transfer(dev_handle, DIRTYJTAG_READ_EP,
+					rx_buf, transfer_length, &actual_length, 1000);
+				if (ret < 0) {
+					cerr << "writeTDI: read: usb bulk read failed " << ret << endl;
+					return EXIT_FAILURE;
+				}
```
trabucayre, line 278: blank before after ';'

src/ftdiJtagBitbang.cpp
```diff
@@ -107,8 +107,7 @@ int FtdiJtagBitBang::setBitmode(uint8_t mode)
 	_bitmode = mode;
 
 	int ret = ftdi_set_bitmode(_ftdi, _tck_pin | _tms_pin | _tdi_pin, _bitmode);
-	ftdi_usb_purge_rx_buffer(_ftdi);
-	ftdi_usb_purge_tx_buffer(_ftdi);
+	ftdi_tcioflush(_ftdi);
```
trabucayre, line 110: `ftdi_tcioflush` is not available with ftdi < 1.5. Use same approach as ftdipp_mpsse.cpp

trabucayre (changes requested):
Some small remarks

src/dirtyJtag.cpp
trabucayre: s/before/after/g

src/dirtyJtag.cpp
phdussud: No. I shouldn't have been there in the first place because the probe side code never reads past the payload of a xfer command.

src/dirtyJtag.cpp
phdussud: Well I increased the working frequency of the device even for the version 1 of the protocol, so it is better to leave it like this I think.

src/ftdiJtagBitbang.cpp
phdussud: Oh... I didn't realize that I left this commit in my PR. Let me fix it.

src/dirtyJtag.cpp
phdussud: fixed, I hope....

trabucayre (approved):
ok

src/dirtyJtag.cpp
trabucayre: Is it backward compatible? When I have added dirtyJtag I was unable to use at an highest frequency (in fact with frequency > 6KHz the probe seems freezed).

phdussud:
The old version of JTAG1 (prior to my first commit) had serious issues with frequencies. Only a few discrete frequencies would work. 600KHz which is the default worked. I recoded the probe using flag testing instead of interrupt and that made it usable for all frequencies (as far as I know).
So in summary, the default frequency is the same as before and works on all versions of the firmware as far as I know.
The newer firmware and JTAG2 will follow the frequency argument up to 1500Khz, where they plateau. The version 3 is supposed to work on devices that support USB high speed bulk transfers and can go higher than 1500Khz. Unfortunately, it is hard to find a breakout board with a processor able to handle USB high speed. Tweensy 4.x boards do but they are expensive.

Patrick

trabucayre:
It's really interesting!
If it's possible to have a firmware 1 with this frequency it's okay. It's just required to add a note/warning to specify the firmware must be build with a recent repository version (and I need to upgrade my probe). It's really important to avoid to introduce regression or breaking support with already available support, or it's required to warn users about that.

phdussud:
It is ok with all versions of the firmware…. If you pass a higher frequency, it gets limited to the previous range by the old firmware. The fact that only a discrete number of frequencies work with the old firmware isn’t addressed here. I think telling people to upgrade is a good idea. They can at least upgrade to the latest DirtyJTAG1 which does not change the pinout for the blue pill. Note that the pinout has changed for the blue pill for DirtyJTAG2 to take advantage of a SPI unit on chip. The other hardware configurations are unchanged but will not be as fast.

Patrick

trabucayre:
A note with the minimal version is required (maybe a subnote about using --freq 600e3 if the update it's not possible may be good too). My probe, with a firmware ~1 year old stall if frequency is >600KHz.
Using last release it's okay for v1.
For v2 with a different pinout I think it's not really a problem since users know which version is used.

phdussud:
Right. I just verified that the old firmware limits the frequency to 1000KHz, which is above the safe limit of 600KHz per your experience.

trabucayre:
Yep, I have limited to 600KHz according to my probe behavior.
Applied, thanks

PR #79, opened 2021-03-24
add support for kintex xc7k325t

GbGp:
add support for kintex xc7k325t, found for example on the KC705 dev board.

While I've not tested it yet, I'm sure flashing the QSPI works as well. Only problem is that a licence in required to generate a bitstream for this device family. So it's probably a good idea to partition the makefile in devices that can be used with the free version, and then the others, or something similar.


trabucayre:
Looks good. Just one remark, since you add a board the README must be updated to list them.

for bitstream generation in `spiOverJtag` it's true, It's a good idea to use an env var to add FPGAs with with a commercial license, and by default bypass these build.
Thanks

trabucayre:
Applied thanks.
I've added the kc705 in supported board list myself.

PR #80, opened 2021-03-27
recognize altera cyclone V 5CEBA4 FPGA

emard:
With this change I can upload *.svf file,
it starts and works

### DIFF
--- a/src/part.hpp
+++ b/src/part.hpp
@@ -27,6 +27,7 @@ static std::map <int, fpga_model> fpga_list = {
 
 	{0x020f30dd, {"altera", "cyclone 10 LP", "10CL025"}},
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Looks good to me, just a remark: since you add a new device/flavor the README must be updated too.

Thanks

emard:
I'm not expert for altera product line but this device seems
identical from programming point of view, only less LUTs

trabucayre:
When I said to update the README, it's just updating cyclone V line with this new code in the *Supported (tested) FPGA:* section

emard:
Aha I guess I should copy exactly what is written on
FPGA chip to let other users know what is really tested

trabucayre:
Applied thanks.
Added this cyclone V model in device supported list myself.

PR #92, opened 2021-06-13
dirtyJtag optimizations to cut the number of USB requests

phdussud:
This is the changes that I mentioned. Right now the speed of dirtyJtag is limited by the USB traffic....

### DIFF
--- a/src/dirtyJtag.cpp
+++ b/src/dirtyJtag.cpp
@@ -179,14 +179,43 @@ int DirtyJtag::setClkFreq(uint32_t clkHZ)
 int DirtyJtag::writeTMS(uint8_t *tms, int len, bool flush_buffer)
 {
 	(void) flush_buffer;
+	int actual_length;
 
 	if (len == 0)
 		return 0;
 
-	for (int i = 0; i < len; i++) {
-		bool val = (tms[i >> 3] & (1 << (i & 0x07)));
-		sendBitBang(SIG_TMS, ((val)?SIG_TMS : 0), NULL, (i == len-1));
+	uint8_t mask= SIG_TCK | SIG_TMS;
+	uint8_t buf[64];
+	u_int buffer_idx = 0;
+	for (int i = 0; i < len; i++)
+	{
+		uint8_t val = (tms[i >> 3] & (1 << (i & 0x07))) ? SIG_TMS : 0;
+		buf[buffer_idx++] = CMD_SETSIG;
+		buf[buffer_idx++] = mask;
+		buf[buffer_idx++] = val;
+		buf[buffer_idx++] = CMD_SETSIG;
+		buf[buffer_idx++] = mask;
+		buf[buffer_idx++] = val | SIG_TCK;
+		if ((buffer_idx + 7) >= sizeof(buf) || (i == len - 1)) {
+			//flush the buffer
+			if (i == len - 1) {
+				//insert tck falling edge
+				buf[buffer_idx++] = CMD_SETSIG;
+				buf[buffer_idx++] = mask;
+				buf[buffer_idx++] = val;
+			}
+			buf[buffer_idx++] = CMD_STOP;
+			int ret = libusb_bulk_transfer(dev_handle, DIRTYJTAG_WRITE_EP,
+										   buf, buffer_idx, &actual_length, 1000);
+			if (ret < 0)
+			{
+				cerr << "toggleClk: usb bulk write failed " << ret << endl;
+				return -EXIT_FAILURE;
+			}
+			buffer_idx = 0;
+		}
 	}
+	assert(buffer_idx == 0);
 	return len;
 }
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Sorry for delays.
It is working for all dirtyJTAG version (I assume we can forget older version since @jeanthom has added a new tag) ?

Comments bellow

src/dirtyJtag.cpp
```diff
@@ -179,14 +179,43 @@ int DirtyJtag::setClkFreq(uint32_t clkHZ)
...
+		buf[buffer_idx++] = val | SIG_TCK;
+		if ((buffer_idx + 7) >= sizeof(buf) || (i == len - 1)) {
+			//flush the buffer
+			if (i == len - 1) {
+				//insert tck falling edge
+				buf[buffer_idx++] = CMD_SETSIG;
+				buf[buffer_idx++] = mask;
+				buf[buffer_idx++] = val;
+			}
+			buf[buffer_idx++] = CMD_STOP;
```
trabucayre, line 207: Maybe buffer overflow risk:
- if at `i == len -2` iteration, you have 8 free Bytes: block 199-206 is not executed;
-  next iteration  will add 6 Bytes (192-198) and  test will match both size and `i == len - 1` conditions
- since it's the last iter the condition in 201 will match: you add 3 Bytes
- finally line 207 you add 1 Bytes
So 10 Bytes are added but only 8 Bytes are available

src/dirtyJtag.cpp
```diff
@@ -179,14 +179,43 @@ int DirtyJtag::setClkFreq(uint32_t clkHZ)
...
+			int ret = libusb_bulk_transfer(dev_handle, DIRTYJTAG_WRITE_EP,
+										   buf, buffer_idx, &actual_length, 1000);
+			if (ret < 0)
+			{
+				cerr << "toggleClk: usb bulk write failed " << ret << endl;
```
trabucayre, line 212: s/toggleClk/writeTMS/g

src/dirtyJtag.cpp
```diff
@@ -179,14 +179,43 @@ int DirtyJtag::setClkFreq(uint32_t clkHZ)
...
+				return -EXIT_FAILURE;
+			}
+			buffer_idx = 0;
+		}
 	}
+	assert(buffer_idx == 0);
```
trabucayre, line 218: I'm not really happy by `assert` since it's throw an exception,  a return value better to avoid a crash. In fact I don't see situation where `buffer_idx` is different to 0 here.

src/dirtyJtag.cpp
```diff
@@ -316,48 +345,52 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
 
 int DirtyJtag::sendBitBang(uint8_t mask, uint8_t val, uint8_t *read, bool last)
```
trabucayre, line 346: With previous `writeTMS` modifications `sendBitBang` method is called once (in `writeTDI`). Maybe it make sense to supress this method and to add required code directly instead of calling this.
Since behaviour is known this will reduce code.

src/dirtyJtag.cpp
phdussud: Sure, I can remove it. I want to point out that assert throws only in debug builds, and that's what debug builds are for. This is a no op for non debug builds, which is the default for this code base.

src/dirtyJtag.cpp
phdussud: Sure

src/dirtyJtag.cpp
phdussud: Great catch! Thanks!

src/dirtyJtag.cpp
```diff
@@ -316,48 +345,52 @@ int DirtyJtag::writeTDI(uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
-				buf, 4, &actual_length, 1000) < 0) {
-			cerr << "sendBitBang usb bulk write failed" << endl;
-			return -EXIT_FAILURE;
 		}
-	}
 
-	return EXIT_SUCCESS;
+		return EXIT_SUCCESS;
+	} else {
+		return toggleClk(SIG_TMS & val, SIG_TDI & val, 1);
```
trabucayre: toggleClk receive tms & tdi state. so by using SIG_TMS and SIG_TDI, both are always different to 0 so tdi will be always set. This line must be
```
return toggleClk(1, last_bit, 1);
```
at this step tms must be high, and tdi depends on last bit.

src/dirtyJtag.cpp
trabucayre: True. Not enough coffee :)

src/dirtyJtag.cpp
trabucayre: seems good. Maybe it's possible to consider required size according to end. But current solution LGTM.
thanks

trabucayre:
LGTM. Just a small fix with toggleClk.

trabucayre:
Applied thanks.
Will fix toggleclk by myself (and some small tricks)

phdussud:
Thank you very much! 
Patrick

PR #94, opened 2021-06-22
Add Orbtrace support

zyp:
Hi, [Orbtrace](https://github.com/orbcode/orbtrace) is an open source FPGA-based debug and trace interface with CMSIS-DAP support. Attached patches makes openFPGALoader work with it.

The second patch removes the product string check. CMSIS-DAP specification states that the string «CMSIS-DAP» can either be placed in the product string or the interface string for the actual interface, and since Orbtrace is a composite device with both CMSIS-DAP v1 (HID) and v2 (bulk) interfaces being optional features, presence is only being announced by each interface string.

Since hidapi doesn't expose interface strings, there's no easy way to check them, and since we're currently only calling `hid_enumerate()` with the VID/PID combination for the selected cable anyway, I propose just removing all string checking for the time being.

To later support arbitrary CMSIS-DAP devices without having to whitelist each VID/PID combination, I suggest using libusb to enumerate devices, checking both product and interface strings, before switching to hidapi to open the interface if a v1 interface is selected.

```
zyp@rin build $ ./openFPGALoader -c orbtrace --detect 
write to ram
Found 1 compatible device:
	0x1209 0x3443 Orbtrace
index 0:
	idcode 0x41112043
	manufacturer lattice
	model  LFE5U-45
	family ECP5
	irlength 8
```

IMG_20210622_232951

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ __Supported cables:__
 * anlogic JTAG adapter
 * [digilent_hs2](https://store.digilentinc.com/jtag-hs2-programming-cable/): jtag programmer cable from digilent
 * [cmsisdap](https://os.mbed.com/docs/mbed-os/v6.11/debug-test/daplink.html): ARM CMSIS DAP protocol interface (hid only)
+* [Orbtrace](https://github.com/orbcode/orbtrace): Open source FPGA-based debug and trace interface
 * [DFU (Device Firmware Upgrade)](http://www.usb.org/developers/docs/devclass_docs/DFU_1.1.pdf): USB device compatible with DFU protocol
 * [DirtyJTAG](https://github.com/jeanthom/DirtyJTAG): JTAG probe firmware for STM32F1
   (Best to use release (1.4 or newer) or limit the --freq to 600000 with older releases. New version https://github.com/jeanthom/DirtyJTAG/tree/dirtyjtag2 is also supported)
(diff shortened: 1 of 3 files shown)


trabucayre:
Hi.
Really interesting project. This is, for me a way to have access to a v2 probe to complete this support.
Instead of removing product string maybe simply adding a Work around to add inconditionaly device with orbtrace VID/PID?

trabucayre:
In fact I've no ECPIX-5 nor the dedicated board. But maybe an ULX3S, an orangeCrab or an tinyFPGA do the job?
Thanks

zyp:
The code currently assumes the hardware got ULPI for HS USB, but making it work with FS USB and testing on ULX3S is somewhere on my TODO list. Haven't been prioritized since FS USB is too slow to be very useful for trace, but it should be good enough for debug.

Since CMSIS-DAP v2 is based on plain bulk, you'll be using libusb to access v2 interfaces, in which case it's easy enough to check the interface strings as well. Once you've found an interface matching the string, you could just check whether `bInterfaceClass` is 0x03(HID) for v1 or 0xff (vendor defined) for v2, and if it's v1 you can then switch to hidapi for opening it.

trabucayre:
Ok. Thank. So I must order an ECPIX-5 to test.
I read a bit about v2, but needs to have compatible hardware to avoid "blind dev" :)

trabucayre:
I assume there is no currently way to order one of your board?

trabucayre:
Thanks for your PR.
When I have compatible hardware I will implement v2 support and maybe some sanity check to avoid using not cmsis-dap compatible devices.

zyp:
> I assume there is no currently way to order one of your board?

We've just got a few initial prototypes to validate the design at the moment, and are currently preparing for the first production run. If you'd like a notification when they are available to order, you can subscribe to orbcode/orbtrace#1.

trabucayre:
Done. Thanks

PR #97, opened 2021-07-03
Adds support for the Xilinx Spartan 6

wasv:
Adds support for writing to SPI flash with the Xilinx Spartan 6.
Tested with the xc6slx100fgg484 used in the Pano Logic G2 with a M25P64 flash chip.


wasv:
This is currently marked as a draft because the process of building the bitstream is not yet added to the Makefile.

trabucayre:
Hi. Thanks to start this feature, it's in my TODO list for a while :-)
I have some remarks:
- You use a subset of FLASH signals,  maybe for your pano logic G2 it's enough but in for some others (as pipistrello) `hold` and `wp` are present so it's better to provides full set to be compatible.
- STARTUPE2 (or something like this) is not required ?
- your tcl script is a dump for ISE, it's usually possible to have something less verbose.

For Makefile, it's clear a better approach must be found. I'm currently working on altera version of this bridge and consecuently this add more complexity again. I think to move to a mix Makefile/edalize to reduce complexity of Makefile/tcl/tool.

trabucayre:
I think you want said spartan instead of cyclone ;)

wasv:
> I think you want said spartan instead of cyclone ;)

You're correct, thanks. This PR was submitted pretty late one night, so there's definitely work left to do on it. Should be able to tidy things up at some point soon.

trabucayre:
It's look better. I need to test.
I have do some try with edalize and It seems a good approach to drastically simplify Makefile and scripts. It's maybe possible by using a generic to have a unique vhdl file for serie 6 and 7 to simplify again this directory.

Thanks

trabucayre:
I've tried with my pipistrello (after some modifications) and It's works. for this board, it's weird since there is pullup resistors, but wpn and hldn must be explictly provided at ucf level.
One last thing, could you rename bit file according to existing ( xc6slx100fgg484) ?
Thanks

trabucayre:
Applied to master
Thanks

PR #98, opened 2021-07-11
Add board specific default frequency

ultraembedded:
As some boards are known to run with higher TCK frequencies, allow a board-specific default to be specified in the board configuration table.

The only board affected by this change is the Digilent Arty which now defaults to 10MHz JTAG clock (which I tested), but the mechanism is generic and other boards can be modified as and when required.

The user can still override the frequency from the command line.

### DIFF
--- a/src/board.hpp
+++ b/src/board.hpp
@@ -80,34 +80,38 @@ typedef struct {
 	uint16_t mode;           /*! communication type (JTAG or SPI) */
 	jtag_pins_conf_t jtag_pins_config; /*! for bitbang, provide struct with pins value */
 	spi_pins_conf_t spi_pins_config; /*! for SPI, provide struct with pins value */
+	uint32_t default_freq;   /* Default clock speed: 0 = use cable default */
 } target_board_t;
 
-#define JTAG_BOARD(_name, _fpga_part, _cable, _rst, _done) \
-	{_name, {"", _cable, _fpga_part, _rst, _done, COMM_JTAG, {}, {}}}
+#define CABLE_DEFAULT 0
+#define CABLE_MHZ(_m) ((_m) * 1000000)
+
+#define JTAG_BOARD(_name, _fpga_part, _cable, _rst, _done, _freq) \
+	{_name, {"", _cable, _fpga_part, _rst, _done, COMM_JTAG, {}, {}, _freq}}
 #define JTAG_BITBANG_BOARD(_name, _fpga_part, _cable, _rst, _done, _tms, _tck, _tdi, _tdo) \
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM. see below for some small remarks

src/board.hpp
```diff
@@ -80,34 +80,38 @@ typedef struct {
...
+
+#define JTAG_BOARD(_name, _fpga_part, _cable, _rst, _done, _freq) \
+	{_name, {"", _cable, _fpga_part, _rst, _done, COMM_JTAG, {}, {}, _freq}}
 #define JTAG_BITBANG_BOARD(_name, _fpga_part, _cable, _rst, _done, _tms, _tck, _tdi, _tdo) \
-	{_name, {"", _cable, _fpga_part, _rst, _done, COMM_JTAG, { _tms, _tck, _tdi, _tdo }, {}}}
+	{_name, {"", _cable, _fpga_part, _rst, _done, COMM_JTAG, { _tms, _tck, _tdi, _tdo }, {}, 0}}
 #define SPI_BOARD(_name, _manufacturer, _cable, _rst, _done, _cs, _sck, _si, _so, _holdn, _wpn) \
 	{_name, {_manufacturer, _cable, "", _rst, _done, COMM_SPI, {}, \
-		{_cs, _sck, _so, _si, _holdn, _wpn}}}
+		{_cs, _sck, _so, _si, _holdn, _wpn}, 0}}
```
trabucayre, line 95: `--freq` is also used for SPi and BITBANG. Seems logical to add a default freq too.

src/main.cpp
```diff
@@ -95,6 +95,8 @@ int main(int argc, char **argv)
 		board = &(board_list[args.board]);
 	}
 
+	uint32_t freq = args.freq;
+
```
trabucayre, line 101: Why not using args.freq everywhere and overwrite it when required ?

src/main.cpp
```diff
@@ -127,7 +129,7 @@ int main(int argc, char **argv)
 			args.fpga_part = board->fpga_part;
 
 		/* Some boards can override the default clock speed */
-		if (board->default_freq != CABLE_DEFAULT)
+		if (board->default_freq != CABLE_DEFAULT && freq == DEFAULT_FREQ)
```
trabucayre, line 132: What happens if the board has a default freq != 0 and user ask for 6MHz?
It's maybe required to set `args.freqs = 0` and, after board test, something like:
```
if (args.freq == 0) // user has not used --freq and board default freq is empty
    args.freq = DEFAULT_FREQ;
```

trabucayre:
Applied to master
Thanks

PR #100, opened 2021-07-17
ci: add Ubuntu jobs

umarcor:
This PR adds Ubuntu jobs to the CI workflow. As commented in #65, it'd be desirable to use a DEB recipe for generating the package, instead of just creating a tarball (which does not have metadata about the runtime dependencies). However, the tarball is useful, since it allows users to get binaries (as long as they install the deps themselves).

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -8,6 +8,100 @@ on:
 jobs:
 
 
+  lin-build:
+    strategy:
+      fail-fast: false
+      max-parallel: 4
+      matrix:
+        os:
+        - 18
+        - 20
+    name: '🚧🐧 Ubuntu ${{ matrix.os }} | build'
+    runs-on: ubuntu-${{ matrix.os }}.04
+    steps:
+
+    - name: '🧰 Checkout'
+      uses: actions/checkout@v2
+
+    - name: '⚙️ Install dependencies'
+      run: |
+        sudo apt update -qq
+        sudo apt install -y \
+          cmake \
+          libftdi1-2 \
+          libftdi1-dev \
+          libhidapi-libusb0 \
+          libhidapi-dev \
+          libudev-dev \
+          pkg-config \
+          tree
+
+    - name: '🚧 Build tarball'
+      run: |
+        mkdir build
+        cd build
+        cmake ..
+        cmake --build .
+
+        make DESTDIR=$(pwd)/dist install
+        _rulesdir='dist/etc/udev/rules.d/'
+        mkdir -p "${_rulesdir}"
+        cp ../99-openfpgaloader.rules "${_rulesdir}"
+        _licenses="dist/usr/local/share/licenses/openFPGALoader"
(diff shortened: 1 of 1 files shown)


.github/workflows/Test.yml
```diff
@@ -8,6 +8,100 @@ on:
...
+          tree
+
+    - name: '🚧 Build tarball'
+      run: |
+        mkdir build
+        cd build
+        cmake ..
+        cmake --build .
+
+        make DESTDIR=$(pwd)/dist install
```
trabucayre, line 46: you can use `-DCMAKE_INSTALL_PREFIX=$(pwd)/dist` it's more cmake compliant

umarcor: I'm not sure it's the same. The prefix needs to be `/usr/local`, because that is where the package is to be installed. `DESTDIR` is just for copying the content into a different location. Some tools do ignore their PREFIX, so it's not a problem if it does not match the final location. However, others need the prefix to be `/usr/local`.

.github/workflows/Test.yml
```diff
@@ -8,6 +8,100 @@ on:
...
+          libhidapi-libusb0 \
+          udev
+
+    - name: '📥 Download artifact: package'
+      uses: actions/download-artifact@v2
+      with:
+        name: ubtuntu${{ matrix.os }}.04-openFPGALoader
+
+    - name: '🛠️ Install package'
+      run: sudo tar -xvzf *.tgz -C /
```
trabucayre, line 97: Since install root directory is `dist` you will have `dist/bin` and `dist/share`.
If you extract archive in `/` binary file will be installed in `/bin/`. It's usually a directory used for system binaries only available for `root`
So maybe it's required to fix `DESTDIR` to `$(pwd)/dist/usr` to install in `/usr/bin` (usually used by packet manager)

trabucayre:
Whoua it's PR number 100 ! \o/

See my remarks in commit.

.github/workflows/Test.yml
trabucayre: I'm wrong by not using `DCMAKE_INSTALL_PREFIX` the install target will copy binary to `$(pwd)/dist/usr/local/bin`

umarcor: Since the PREFIX is `/usr/local` by default, the tarball does contain that hierarchy (see https://github.com/trabucayre/openFPGALoader/runs/3093795863?check_suite_focus=true#step:6:7). So, `make DESTDIR=dist install` creates the whole hierarchy of directories: DESTDIR/PREFIX/bin and DESTDIR/PREFIX/share.

Then, when extracting it to the root, the content is copied to `/usr/local/bin`, `/usr/local/share` and `/etc/udev/rules.d/`.

trabucayre: Yep. My fault. I've done a mistake

trabucayre:
Applied.
Thanks

PR #101, opened 2021-07-17
ci: use eine/tip to release artifacts

umarcor:
Artifacts don't have a fixed URL and users need to login into GitHub in order to download them. Conversely, release assets do have a fixed URL and can be downloaded without login. Hence, this PR uses an action ([eine/tip](https://github.com/eine/tip)) for uploading the artifacts as assets of pre-release 'nightly'.

Note that it will only upload them if all jobs are successful, and only on branch 'master' or on tagged commits.

Actually, it works for tagged releases too. On untagged commits, the assets in 'nightly' will be updated. On tagged commits, a release will be created; it uses semver to decide whether it needs to be a release or a pre-release.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -194,3 +194,24 @@ jobs:
       run: |
         openFPGALoader --help
         openFPGALoader --detect || true
+
+
+  Release:
+    if: github.event_name != 'pull_request' && (github.ref == 'refs/heads/master' || contains(github.ref, 'refs/tags/'))
+    needs: [ lin-test, win-test ]
+    runs-on: ubuntu-latest
+    name: '📦 Release'
+    steps:
+
+    - name: '📥 Download artifacts'
+      uses: actions/download-artifact@v2
+      with:
+        path: artifact
+
+    # Tagged: create a pre-release or a release (semver)
+    # Untagged: update the assets of pre-release 'nightly'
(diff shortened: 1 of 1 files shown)


trabucayre:
Thanks for this PR.
For 'nightly' this adds a new release each day or update the future official release?

umarcor:
> For 'nightly' this adds a new release each day or update the future official release?

It creates a 'nightly' pre-release the first time, in case it does not exist. Then, it updates the assets in that same pre-release and it force-pushes the nightly tag so that it points to the last successfully run commit. As you can see in https://github.com/ghdl/ghdl/releases, we used eine/tip in GHDL first on 21 May 2020. Since then, the title, the assets and the tag are automatically updated after each successful CI run of an untagged commit.

Removing and creating the pre-release again is avoided so that watchers are not continuously notified about a new release.

This is not a "future official release". When you push a tagged commit, the 'nightly' is ignored. It will create a new release matching the name of the tag and it will upload the artifacts there. Whether it's a release or a pre-release depends on the semver you use in the tag. The next commits you push which are not tagged will use the existing 'nightly' again.

Overall, the process of gathering the artifacts and uploading them as the assets of a (pre)release is always the same. The difference is the name of the target (pre)release, which depends on the commit being tagged or not.

trabucayre:
Good feature.
Applied 
Thanks

PR #112, opened 2021-09-07
doc: retrieve username from $USER

wuhanstudio:
Hi,

This PR adds a small modification to the documentation, so that we don't need to replace YourUserName manually.


trabucayre:
Good idea!
Thanks

PR #113, opened 2021-09-20
progressBar: Use only stdout.

zyp:
The progress markers of the progress bar were output to stderr while the rest was output to stdout.

This patch moves everything to stdout.


trabucayre:
Good catch!
Thanks

PR #122, opened 2021-10-03
CMakeLists.txt: Default to USEUDEV=OFF in windows builds.

UweBonnes:


trabucayre:
Good remark. But maybe simply `set(ENABLE_UDEV OFF)` for windows make more sense

UweBonnes:
Changed

trabucayre:
Applied.
Thanks

PR #126, opened 2021-10-14
small contrib for ISSI 16MB protection

emard:
Here's added protection descriptor for 16MB ISSI,
the datasheet is the same as 8MB and 4MB ISSI.

I have 16MB IS25LP128 and 4MB IS25LP032

The 4MB seems to have bug, although Top or Bottom
bit can be selected, the chip always protects Top
and seems to ignore this bit.

fixed also small typo windbond->winbond


emard:
Yes - in the descriptor database there can be
2 byte of commands: to read and to write each status
register. With bitmask of bit meaning, it would cover
2 mention chip types we have tested

trabucayre:
Current db is more or less a proof of concept: I try in a first time to focus on most common chip's behaviour to have a first working base. And in second time, improving model with more specific use case.

PR #134, opened 2021-11-25
Add bitstream authentication for MachXO3D

abacomartin:
Added programming function to program, verify the public key for bitstream authentication - also enables authentication


trabucayre:
LGTM!
Applied thanks.
It's a note to me: quite all `wr_rd` are followed by `_jtag->set_state(Jtag::RUN_TEST_IDLE);`. Needs to check the few exceptions to that and to move directly to `RUN_TEST_IDLE` into `wr_rd`

PR #139, opened 2021-12-01
ci: rm 'max-parallel'

umarcor:
Since this repo is public and open source, there is no need to limit the maximum number of parallel jobs. If more jobs are required, additional jobs will be queued and executed automatically. Nonetheless, the limit is 18-20 concurrent jobs, which is not frequent in this repo.


trabucayre:
Applied!
Thanks

PR #140, opened 2021-12-01
ci/win: use option 'pacboy' of setup-msys2

umarcor:
In v2.6.0 of Action setup-msys2, option 'pacboy' was added. That allows to slightly cleanup the syntax used in the CI workflow.


trabucayre:
Applied.
Thanks!

PR #141, opened 2021-12-01
ci/win: add job 'win-jobs'

umarcor:
Close #140 

This PR is based on #140. Instead of declaring the list of MSYS2 jobs twice, it is declared once, and used in the matrices of jobs win-makepkg and make-test.


trabucayre:
Thanks!

PR #143, opened 2021-12-01
ci/Release: replace openFPGALoader-Documentation subdir with a tarball

umarcor:
When the Sphinx documentation was added, I overlooked that the whole site was being uploaded to the "nightly" release, uncompressed! I'm sorry about that 😭 

This PR fixes that issue. The documentation artifact is named, and it is compressed in a tarball, so that the release action can push it without file explosion.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -259,4 +259,4 @@ jobs:
       with:
         token: ${{ github.token }}
         tag: 'nightly'
-        files: artifact/*-openFPGALoader/*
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied, thanks!
I have definitively to read books/docs about github ;-)

umarcor:
This summary might be helpful: https://github.com/pyTooling/Actions/#context

Nonetheless, do not hesitate to ask. There are several non intuitive caveats.

PR #145, opened 2021-12-05
Update install.rst

RGD2:
Build instructions from source for debian-based distros fixed

- apt install line was broken, missing a `\`, 
- a git clone line was added to clarify when and where source should be
- Build directory also created beside tree rather than inside, and the cmake line fixed to match
- tested on Debian GNU/Linux 11 (bullseye) / WSL2

### DIFF
--- a/doc/guide/install.rst
+++ b/doc/guide/install.rst
@@ -32,7 +32,7 @@ This application uses ``libftdi1``, so this library must be installed (and, depe
 
 .. code-block:: bash
 
-    apt-get install
+    sudo apt install \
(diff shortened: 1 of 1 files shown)


RGD2:
Just a couple fewer things for newbies to trip over.

trabucayre:
Good catch!
It's just a bit unusual to create **build** outside tree. Each time I read install doc for something CMake based it's inside, and it's make sense to avoid weird behaviour in some situation

trabucayre:
Applied manually with requested changes.
Thanks

PR #147, opened 2021-12-10
part: add support for Anlogic ELF2 EF2M45

Icenowy:
Anlogic EF2M45 is a FPGA with a co-packaged 4Mbit SPI Flash, and the
JTAG interface is the same with EG4S20.

Add support for it by adding it to the part database.

SPI Flash programming and SRAM programming are both tested.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

### DIFF
--- a/doc/compatibility/fpga.rst
+++ b/doc/compatibility/fpga.rst
@@ -7,6 +7,7 @@ FPGAs
  Vendor  Model                                                                                                                               Memory Flash
 ======== =================================================================================================================================== ====== =====
 Anlogic  `EG4S20 <http://www.anlogic.com/prod_view.aspx?TypeId=10&Id=168&FId=t3:10:3>`__                                                     OK     AS
+Anlogic  `EF2M45 <http://www.anlogic.com/prod_view.aspx?TypeId=12&Id=170&FId=t3:12:3>`__                                                     OK     AS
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM
But could you update [fpga list](https://github.com/trabucayre/openFPGALoader/blob/master/doc/compatibility/fpga.rst) too?
Are there a board based on this device ?
Thanks

Icenowy:
No board is commercially available now, as I know. (BTW EF2L45LG144B has MCU enabled too, so it should be in EF2M45 family.  This may have some boards available)

I designed one with a few samples from Anlogic, and open-sourced it (current version on my GitHub is a broken one from 2019, I reworked it some days ago, waiting for JLCPCB to deliver the new PCB).

Icenowy:
BTW for the FPGA list, EG4S20 should be Flash OK, not AS (because it's still flashing via JTAG, although a JTAG-SPI channel is built into Anlogic FPGA hardware, so it's why EF2M45 internal flash can be flashed)

trabucayre:
It's true: flash status must be OK instead of AS -> fixed.
Feel free to add your board if everything is available to reproduce / have access. The honeycomb is referenced but there is no store to bough one.

trabucayre:
Applied to master!
Thanks

PR #148, opened 2021-12-11
Cologne Chip GateMate integration

pu-cc:
This commit adds support for the Cologne Chip GateMate FPGA series.

Both Evaluation Board and Programmer Cable are supported. Configurations can be loaded into the FPGA with both devices via JTAG or SPI. In addition to reading/writing data from/to flashes directly via SPI, this can also be done via the built-in JTAG-SPI-bypass. A direct wiring between programming hardware and flash is no longer necessary in this case.

* Added part with JTAG ID to `src/part.hpp`. We keep the IDCODEs highest nibble to prevent confusion with Efinix T4/T8.
* Added programmer and board with its relevant information to `src/cable.hpp` and `src/board.hpp`.
* Check the IDCODEs highest nibble in `src/jtag`.
* Make `FTDIpp_MPSSE` a public parent for class `FtdiJtagMPSSE` so that we can access GPIO functionality.
* `src/colognechipCfgParser.{cpp,hpp}` contains the ASCII configuration parser.
* `src/colognechip.{cpp,hpp}` contains the basic functionality to write the configuration to FPGA or access external flashes via JTAG/SPI.
* Usage is described in `doc/vendors/colognechip.rst`.

**Connections of the `FT232H`-based programming cable:**
(`ADBUS3..0` configurable for SPI or JTAG communication)

image

**Connections of the `FT2232H`-based evaluation board:**
(`ADBUS3..0` is JTAG and `BDBUS3..0` is SPI)

image

Note that the pin configuration for `~OE`, `RST`, `~FAILED` and `DONE` are valid for both evaluation board and programmer cable. I will link the official datasheets here as soon as they become available.


src/colognechip.hpp
```diff
@@ -0,0 +1,62 @@
...
+/*
+ * Copyright (C) 2021 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
+ * Copyright (C) 2021 Cologne Chip AG <support@colognechip.com>
+ */
+
+#ifndef SRC_COLOGNECHIP_HPP_
+#define SRC_COLOGNECHIP_HPP_
+
+#include <unistd.h>
+#include <regex>
```
trabucayre, line 11: missing #include <string>

src/part.hpp
```diff
@@ -107,6 +107,8 @@ static std::map <int, fpga_model> fpga_list = {
 	{0x0100381B, {"Gowin", "GW1N", "GW1N-4", 8}},
 	{0x0300181b, {"Gowin", "GW1NS", "GW1NS-2C", 8}},
 	{0x0100981b, {"Gowin", "GW1NSR", "GW1NSR-4C", 8}},
+
+	{0x20000001, {"colognechip", "GateMate Series", "GM1Ax", 6}},
```
trabucayre, line 111: a note may required to explain why this device has version not masked. It's explained in `jtag.cpp`  but it's useful here to for contributors.

src/colognechipCfgParser.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include "configBitstreamParser.hpp"
+
+class CologneChipCfgParser: public ConfigBitstreamParser {
+	public:
+		CologneChipCfgParser(const std::string &filename, bool reverseOrder);
+
+		int parse() override;
+
+	private:
+		bool _reverseOrder;
```
trabucayre, line 22: `_reverseOrder` is never used. Same remark for constructor

src/colognechip.cpp
```diff
@@ -0,0 +1,427 @@
...
+	uint16_t rstn_pin, uint16_t done_pin, uint16_t failn_pin, uint16_t oen_pin,
+	bool verify, int8_t verbose) :
+	Device(NULL, filename, file_type, verify, verbose), _rstn_pin(rstn_pin),
+		_done_pin(done_pin), _failn_pin(failn_pin), _oen_pin(oen_pin)
+{
+	_spi = spi;
+	_spi->gpio_set_input(_done_pin | _failn_pin);
+	_spi->gpio_set_output(_rstn_pin | _oen_pin);
+
+	if (prg_type == Device::WR_SRAM) {
```
trabucayre, line 24: maybe file type check: doesn't make sense to load raw data in SRAM ?

src/colognechip.cpp
```diff
@@ -0,0 +1,427 @@
...
+	_failn_pin = DBUS6;
+	_oen_pin   = spi_board->oe_pin;
+
+	/* cast _jtag->_jtag from JtagInterface to FtdiJtagMPSSE to access GPIO */
+	_ftdi_jtag = reinterpret_cast<FtdiJtagMPSSE *>(_jtag->_jtag);
+
+	_ftdi_jtag->gpio_set_input(_done_pin | _failn_pin);
+	_ftdi_jtag->gpio_set_output(_rstn_pin | _oen_pin);
+
+	if (prg_type == Device::WR_SRAM) {
```
trabucayre, line 60: file type check again?

trabucayre:
LGTM!
just a small question: you use cfgDone(), with a loop and a timeout, in more: maybe a function like `waitCfgDone` make sense to factorise this piece of code ?

thanks!

src/colognechip.cpp
pu-cc: It actually makes sense, because the P&R can generate the configuration as a bit file (raw) or as ASCII. In the latter, one byte is written out as a hex value in each line. This makes it easier to edit the configuration by hand and e.g. add comments.

The ASCII format would like like this, for example:

```
// Config Command Path   CMD_PATH
// ---------------------------------
D9 // Command: CMD_PATH      addr: 32'h0000 0000
01 // Length: 1
ED // -- CRC low byte
96 // -- CRC high byte
10 // _prog 
4D // -- CRC low byte
D6 // -- CRC high byte
00 // NOP     1
// ... 
```
Since they have the same content, they can both be written to the FPGA latches or to the external flash.

trabucayre: When I talk about raw data it's for example a firmware for a softcore.
The idea is to check if the file is an allowed bitstream and fails otherwise.
Sorry for my imprecise sentence.

pu-cc: Ok now I understand, thank you. The `_file_extension` is checked in `program()`. If you don't mind I would keep the check there, but also handle unknown types (e.g. for firmware):

```cpp
if (_file_extension == "cfg") {
	cfg = new CologneChipCfgParser(_filename);
} else if (_file_extension == "bit") {
	cfg = new RawParser(_filename, false);
} else { /* unknown type: */
	if (_mode == Device::FLASH_MODE) {
		cfg = new RawParser(_filename, false);
	} else {
		throw std::runtime_error("incompatible file format");
	}
}
```

In other words: Use `CologneChipCfgParser` for `*.cfg` and `RawParser` for `*.bit` config files. If the file type is unknown, we only allow writing to flash. Are you ok with that? Otherwise I like to hear your suggestion.

src/colognechip.cpp
```diff
@@ -84,7 +84,7 @@ void CologneChip::reset()
  * Obtain CFG_DONE and ~CFG_FAILED signals. Configuration is successfull iff
  * CFG_DONE=true and ~CFG_FAILED=false.
  */
-bool CologneChip::cfgDone()
+bool CologneChip::waitCfgDone()
```
trabucayre, line 87: My first idea with `cfgDone` and `waitCfgDone` is to left `cfgDone` unchanged but introducing `waitCfgDone` method with the loop, call, timeout and a return `true` or `false`.
This avoid to have same piece of code multiple time.

src/colognechip.cpp
trabucayre: Perfect.

src/colognechip.cpp
pu-cc: You're right this is much better :)

trabucayre:
LGTM (again!)
I just see I have forgotten doc: You need to update:
- compatibility/board.rst, compatibility/cable.rst, compatibility/fpga.rst 
-  README.md and doc/index.rst to add new doc/vendors/colognechip.rst entry

Sorry to missed this before
Thanks

pu-cc:
Sure, that was still missing. I just added this - to the best of my knowledge - and hope I didn't break anything in the >build the docs< section...

doc/compatibility/cable.rst
```diff
@@ -20,3 +20,4 @@ Cables
 * `Tang Nano 4k USB-JTAG interface <https://github.com/sipeed/RV-Debugger-BL702>`__: USB-JTAG/UART debugger based on BL702 microcontroler.
 * `Tigard <https://www.crowdsupply.com/securinghw/tigard>`__: SWD/JTAG/UART/SPI programmer based on Ftdi FT2232HQ
 * `honeycomb USB-JTAG interface <https://github.com/Disasm/f042-ftdi>`__: FT2232C clone based on STM32F042 microcontroler
+* `Cologne Chip GateMate FPGA Programmer <https://colognechip.com/programmable-logic/gatemate/>`__ FT232H-based JTAG/SPI programmer cable
```
trabucayre: missing ':'

trabucayre:
Seems ok!
But your branch has to be rebased: there is conflict for fpga.rst

pu-cc:
Thanks, should be fine now

trabucayre:
Applied to master!
Thanks!

PR #153, opened 2021-12-18
Fix printf data type

uis246:


trabucayre:
Good catch.
Applied to master
Thanks

uis246:
I should mark this as draft. True format even not %lu, but %zu. Also %zd on lines 171 and 378

trabucayre:
Maybe I have merged this PR too late.
Feel free to open a new one to fix printf.
Thanks!

PR #155, opened 2021-12-23
colognechip integration: create CologneChip SPI object in dynamic fashion as well

pu-cc:
This PR adds the missing object creation in for the `CologneChip` class in the new dynamic fashion for SPI targets.

I  tested the master branch on the GateMate programmer and evaluation board hardware and it is OK.


trabucayre:
Good catch!
Thanks!

PR #156, opened 2021-12-30
ci: fix step 'Show package content'

umarcor:
It seems that some change in `tar` is producing issues when trying to extract `.zst` files in the CI workflow. This PR uses `zstd` explicitly, in order to work around it.

I reported the issue to MSYS2 maintainers, and they are having a look at the problem. Hence, this fix might not be needed in a few days. @trabucayre, it's up to you to merge this PR, or to just wait for it to be fixed upstream. The main motivation for this PR is making you aware.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -196,7 +196,8 @@ jobs:
       run: |
         mkdir tmp
         cd tmp
-        tar -xf ../scripts/msys2/*.zst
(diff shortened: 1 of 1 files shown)


trabucayre:
Thanks!
I have seen tar.exe not able to handle zstd files but did not think by myself to this solution.
I have applied your PR. Maybe I whill revert this when issue will be fixed or leave as it since this is acceptable solution to uncompress these files.

PR #157, opened 2022-01-05
ci: use reusable composite action and reusable workflow from msys2/setup-msys2

umarcor:
I contributed a composite action and a reusable workflow to msys2/setup-msys2, in order to reduce the duplication in all the repos building and testing MSYS2 packages. This PR replaces the `win` jobs in the CI workflow in order to use those. The functionality is the same.

Find references about the differences of Action and Workflow kinds in https://github.com/pyTooling/Actions/#context.


trabucayre:
Applied Thanks!

Issue #105, opened 2021-08-11
Join forces with hdl/constraints


umarcor:
*Moved from https://github.com/trabucayre/openFPGALoader/pull/102#issuecomment-895953900.*

@trabucayre, I would like to propose joining forces with [hdl/constraints](https://github.com/hdl/constraints). Subdir [board](https://github.com/hdl/constraints/tree/main/board) contains a subdir for each board. There, an `info.yml` file contains the data about the board (label, device, package, part number, docs, programmers, etc.). In some cases, a `README.md` containing a frontmatter is used instead of the YAML file. See, for instance:

- https://github.com/hdl/constraints/blob/main/board/AC701/README.md
- https://github.com/hdl/constraints/blob/main/board/OrangeCrab/info.yml

Apart from the data, constraint files are provided (in XDC, PCF, LPF, etc.).

Some boards are in the [.todo](https://github.com/hdl/constraints/tree/main/board/.todo) subdir, which means that the constraints were not gathered/reworked yet. However, all of them contain data at least.

The content of hdl/constraints is shown in [hdl.github.io/awesome/boards](https://hdl.github.io/awesome/boards/) (all boards, "todo" and "not todo"). In the list, the board, the device and a picture are shown, along with icons/references to the constraints repo, the datasheets, git repos, etc. There is a page for each board, where the complete `info.yml`/`README.md` content is shown. For instance:

- https://hdl.github.io/awesome/boards/ac701/
- https://hdl.github.io/awesome/boards/orangecrab/

Overall, the main purpose of hdl/constraints is for projects such as openFPGALoader to avoid the maintenance burden of gathering all the references. You can use a single link (`hdl.github.io/awesome/boards/BOARDIDENTIFIER`) to access an "entrypoint" containing multiple references for that board. Moreover, we are discussing about generating (interactive) pinout diagrams (maybe automatically from KiCAD sources): https://github.com/hdl/constraints/issues/3. Then, the page of each board would contain the pinout, apart from the photo/render.

Therefore, I would be willing to adapt/modify the board info template (https://github.com/hdl/constraints/blob/main/template/board.info.yml) in order to better fit the information about openFPGALoader. Say:

```yml
prog?:
  - tool: openFPGALoader
    keyword: <WORD>
    memory: <BOOLEAN>
    flash: <BOOLEAN>
  - <TOOL2>
  - <TOOL3>
```

[...]

PR #160, opened 2022-01-12
doc: declare board compatibility through a YAML file

umarcor:
Instead of manually writing the board compatibility table in the docs, this PR uses a YAML file and a Python script:

- Boards are declared in the YAML file, which provides a machine-readable solution for consumers of this repo to query the data.
- The Python script reads the YAML file and generates a restructuredtext table through [tabulate](https://pypi.org/project/tabulate/).

No content is changed, so the documentation looks exactly the same. This is an internal enhancement related to #105.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,5 +39,7 @@ openFPGALoader
 build/
 
 # Documentation
-/doc/_build
-/doc/_theme
+/doc/__pycache__/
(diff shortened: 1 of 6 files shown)


trabucayre:
Awesome!
Thanks for this update!
Applied!

PR #161, opened 2022-01-12
doc: add field 'Constraints' to 'boards.yml'

umarcor:
This is a follow-up of #160, related to #105.

An optional field, `Constraints`, is added to each item in `boards.yml`. `Constraints` can contain an string or a list of strings, matching the boad identifiers used in [hdl/constraints](https://github.com/hdl/constraints).

At the same time, an intersphinx mapping to hdl/constraints was added to the documentation configuration. When building the documentation, a new column is added to the boards compatibility table, which contains cross-references to [hdl.github.io/constraints/Data/Boards](https://hdl.github.io/constraints/Data/Boards). See [umarcor.github.io/openFPGALoader/compatibility/board](https://umarcor.github.io/openFPGALoader/compatibility/board).

@trabucayre, as discussed in https://github.com/trabucayre/openFPGALoader/issues/105#issuecomment-899657329, the purpose is each repo to can keep their own data/details. These "keywords" allow cross-referencing through interpshinx, as shown in this PR. Furthermore, after this PR is merged, we can use the `boards.yml` and `boards.py` files from this repo to enhance the documentation of hdl/constraints.

We can later improve this iteratively:

- Add misssing boards to hdl/constraints.
- Evaluate adding fields to the YAML here or there, and/or moving fields from one to the other.
- Update the layour in the site here, in order to show a section per board, instead of or apart from the table.
- Etc.

### DIFF
--- a/doc/boards.py
+++ b/doc/boards.py
@@ -15,6 +15,7 @@ class Board:
     FPGA: str = None
     Memory: str = None
     Flash: str = None
+    Constraints: str = None
 
 
 def ReadDataFromYAML():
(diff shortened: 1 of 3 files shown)


trabucayre:
Really useful informations/field!
I have to see how to limit columns size to avoid a big horizontal scroll.
Thanks!
Applied

PR #166, opened 2022-01-21
Add support for Gowin GW1NR-9C

Icenowy:
The chip used by Sipeed Tang Nano 9K, GW1NR-LV9QN88PC6/I5, belongs to GW1NR-9C family.

Add support for it.

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -50,6 +50,7 @@ Gowin:
      - GW1N-1
      - GW1N-4
      - GW1NR-9
+     - GW1NR-9C
      - GW1NS-2C
      - GW1NSR-4C
     URL: https://www.gowinsemi.com/en/product/detail/2/
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -366,6 +366,13 @@
   Memory: OK
   Flash: IF/EF
 
+- ID: tangnano4k
+  Description: Sipeed Tang Nano 9K
+  URL: https://tangnano.sipeed.com/en/
+  FPGA: littleBee GW1NR-9C
+  Memory: OK
+  Flash: IF/EF
+
 - ID: tec0117
   Description: Trenz Gowin LittleBee (TEC0117)
   URL: https://shop.trenz-electronic.de/en/TEC0117-01-FPGA-Module-with-GOWIN-LittleBee-and-8-MByte-internal-SDRAM
--- a/src/board.hpp
+++ b/src/board.hpp
@@ -158,6 +158,7 @@ static std::map <std::string, target_board_t> board_list = {
 	JTAG_BOARD("runber",          "", "ft232",      0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("tangnano",        "", "ch552_jtag", 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("tangnano4k",      "", "ft2232",     0, 0, CABLE_DEFAULT),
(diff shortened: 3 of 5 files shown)


Icenowy:
This is not a hack -- this is a fix. The current problem is that the idcode read from the real hardware (which isn't under our control) is not masked.

In addition, the idcode added in the first commit it is exactly the same with the fs file generated by Gowin toolchain when selecting GW1NR-9C, despite after masking version nibble it's the same with GW1N-6, not GW1N-9.

The part detection code in main.cpp doesn't take version nibble into account now either. 

As reference, the following is copied from UG290-2.4 (currently available at http://cdn.gowinsemi.com.cn/UG290.pdf ):

```
GW1N-1S h09003 h0900381B
GW1NZ-1 h01006 h0100681B
GW1NS-2 h03000 h0300081B
GW1NS(R)-2C h03001 h0300181B
GW1NSE-2C h03001 h0300181B
GW1N-2/2B h01206 h0120681B
GW1N-1P5/1P5B h01206 h0120681B
GW1N(R)-4 h01001 h0100381B
GW1N(R)-4B h11003 h1100381B
GW1N(R)-4C h11003 h0100181B
GW1NS(ER)-4C h01009 h0100981B
GW1N(R)-9 h11005 h1100581B
GW1N(R)-9C h11005 h1100481B
GW2A(R)-18/18C h00000 h0000081B
GW2A-55/55C h00002 h0000281B
```

(Well the data here is more strange, and have even some inconsistency, and even no GW1N-6 here)

Icenowy:
BTW they now do not differentiate GW1N-9 with GW1N-9C on their website's product line page, but they're still different in Gowin IDE V1.9.8.02.

(And GW1N-6's existence is also old data)

Considering no version-agnostic match is implemented in main.cpp now, listing 0x1100481b as GW1N(R)-9C's idcode will just make everything work (and at least we should add an item for 9C) and it exactly matches what the vendor currently says.

trabucayre:
In fact idcode  used in gowin.cpp (and in main.cpp) is retrieved [here](https://github.com/trabucayre/openFPGALoader/blob/master/src/gowin.cpp#L85) the `get_target_device_id` is filled [here](https://github.com/trabucayre/openFPGALoader/blob/master/src/jtag.cpp#L133) with or without masking high nibble according to part.hpp list.
But since gowin devices may not be differentiate with nibble masked it make sense to replace `_jtag->get_target_device_id` by `idCode()` call to have the full and real idcode without any mask (as done for [lattice device](https://github.com/trabucayre/openFPGALoader/blob/master/src/lattice.cpp#L622)) and dropping mask applied to the idcode read from the file.

Icenowy:
BTW for GW1N-6 I realized that I saw it in fsparser.cpp file. (I thought I have seen some GW1N-6 with its idcode same with GW1N-9C except version nibble). Well there's in fact no such thing known as GW1N-6 now.

And I saw in some old document from Gowin that 0x0100581b is ES of GW1N-9 (and 0x1100581b is final version).

So the version number should still be the not significant part (we can assume GW1N-6 is just a pending name of GW1N-9C when developing).

Icenowy:
The newest pushed version assumes GW1N-6 as GW1N-9C(ES).

trabucayre:
LGTM!
Just a last remark: you need to update [doc/FPGA.yml](https://github.com/trabucayre/openFPGALoader/blob/master/doc/FPGAs.yml#L46) with this model.
And feel free to add the sipeed tang nano 9k ;-)
Thanks

trabucayre:
Applied to master!
Thanks!

PR #171, opened 2022-01-27
board: add tangnano1k

chiplet:
Adds support for the [GW1NZ-1](https://www.gowinsemi.com/en/product/detail/46/) chip and the new [Tang Nano 1K](https://www.aliexpress.com/item/1005002551785169.html?spm=a2g0o.store_pc_groupList.8148356.2.78981d42dwD89o) board.

I'm not entirely sure about what's the purpose of the `Memory` and `Flash` fields in the `boards.yml` file (copy-pasted from `tangnano4k`) so they might not be set correctly.

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -53,6 +53,7 @@ Gowin:
      - GW1NR-9C
      - GW1NS-2C
      - GW1NSR-4C
+     - GW1NZ-1
     URL: https://www.gowinsemi.com/en/product/detail/2/
     Memory: OK
     Flash: IF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -372,7 +372,7 @@
   URL: https://tangnano.sipeed.com/en/
   FPGA: littleBee GW1NZ-1
   Memory: OK
-  Flash: IF/EF
+  Flash: IF
 
 - ID: tangnano4k
   Description: Sipeed Tang Nano 4K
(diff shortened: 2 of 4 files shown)


trabucayre:
LGTM, but you need to update board.hpp to add an entry for this board to be able to use something like `-b tangnano1k`
IF mean internal flash, EF external flash. If you have tried only to load a bitstream in ram, or to flash bitstream into internal flash it's maybe better to drop EF.
Thanks!

chiplet:
I haven't been able to load any bitstreams, since GW1NZ-1 isn't yet supported by the toolchain I'm planning to use:
https://github.com/YosysHQ/apicula/issues/79

trabucayre:
Do you prefer I postpone integration until you have tested, or I apply this and fix if need.
Anyway I have to order 9K and 1K to do a complete check of full functionalities
Thanks

chiplet:
I'd prefer postponing integration until this has been tested to be working with at least some bitstream.

chiplet:
I went ahead and tested the [example designs](https://github.com/sipeed/TangNano-1K-examples) using the official Gowin toolchain. Flashing to both memory and internal flash work correctly. The board doesn't actually even have an external flash chip so I've dropped that from the doc as well. I think this should be ready to merge now.

trabucayre:
Great!
True, if no external flash it make sense.

Applied
Thanks!

trabucayre:
Great!
True, if no external flash it make sense.

Applied
Thanks!

PR #173, opened 2022-02-08
Nicer layout for the boards/fpga/cables table

rstephan:
A little bit cosmetics.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -784,13 +784,13 @@ void displaySupported(const struct arguments &args)
 {
 	if (args.list_cables == true) {
 		stringstream t;
-		t << setw(15) << left << "cable name:" << "vid:pid";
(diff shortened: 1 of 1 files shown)


trabucayre:
LGTM thanks!
But I see a small issue with cables. vid pid are completed with serie of 0 at right size instead of left:
0x2b become 0x2b00 instead of 0x002b and it's the same for 0x403 becoming 0x4030.

rstephan:
Damned! I do more `printf` then `cout`. ;-)
Thanks for the advice.

trabucayre:
LGTM.
Applied thanks!
Yep `cout` is a bit more boring for that. I have, today, modified a message and replaced `cout` by `snprintf` but maybe I must revert to `cout` to be more cpp compliant :)

Issue #159, opened 2022-01-11
Support for Terasic DE1-SoC board with Cyclone V


sedwards-lab:
The Terasic DE1-SoC board has a Cyclone V on it. Specifically, a 5CSEMA5F31C6 connected through an onboard usb-blasterII

Here's the two-line patch to add support for this board:

```
diff --git a/src/board.hpp b/src/board.hpp
index 6750ec0..fefd67e 100644
--- a/src/board.hpp
+++ b/src/board.hpp
@@ -122,6 +122,7 @@ static std::map <std::string, target_board_t> board_list = {
        JTAG_BOARD("de0",             "", "usb-blaster",0, 0, CABLE_DEFAULT),
        JTAG_BOARD("de0nano",         "ep4ce2217", "usb-blaster",0, 0, CABLE_DEFAULT),
        JTAG_BOARD("de0nanoSoc",      "", "usb-blasterII",0, 0, CABLE_DEFAULT),
+       JTAG_BOARD("de1Soc",          "5CSEMA5", "usb-blasterII",0, 0, CABLE_DEFAULT),
        JTAG_BOARD("de10nano",        "", "usb-blasterII",0, 0, CABLE_DEFAULT),
        JTAG_BOARD("ecp5_evn",        "", "ft2232",     0, 0, CABLE_DEFAULT),
        SPI_BOARD("fireant",              "efinix", "ft232",
diff --git a/src/part.hpp b/src/part.hpp
index 55dc6cd..c9934c5 100644
--- a/src/part.hpp
+++ b/src/part.hpp
@@ -60,6 +60,7 @@ static std::map <int, fpga_model> fpga_list = {
        {0x02b150dd, {"altera", "cyclone V", "5CEA2", 10}},
        {0x02b050dd, {"altera", "cyclone V", "5CEBA4", 10}},
        {0x02d020dd, {"altera", "cyclone V Soc", "5CSEBA6", 10}},
+       {0x02d120dd, {"altera", "cyclone V Soc", "5CSEMA5", 10}},
        {0x02d010dd, {"altera", "cyclone V Soc", "5CSEMA4", 10}},
 
        {0x00000001, {"efinix", "Trion",    "T4/T8",            4}},
```

Incidentally, the FPGA is second on the JTAG chain, which the tool seems to handle correctly:

```
USB-Blaster II firmware version: 1.39
index 0:
	idcode   0xba00477
	type     ARM cortex A9
	irlength 4
index 1:
[...]

PR #174, opened 2022-02-09
Add support  for Terasic DE1-SoC board

tarikgraba:
This pull request should be attributed to @sedwards-lab (see issue #159)

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -81,6 +81,15 @@ Intel:
     Memory: OK
     Flash: OK
 
+  - Description: Cyclone V SE SoC
+    Model:
+      - 5CSEBA6
+      - 5CSEMA4
+      - 5CSEMA5
+    URL: https://www.intel.com/content/www/us/en/products/programmable/fpga/cyclone-v.html
+    Memory: OK
+    Flash: NT
+
   - Description: Cyclone 10 LP
     Model: 10CL025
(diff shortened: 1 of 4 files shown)


trabucayre:
LGTM but please update [FPGA list](https://github.com/trabucayre/openFPGALoader/blob/master/doc/FPGAs.yml#L62) and [board list](https://github.com/trabucayre/openFPGALoader/blob/master/doc/boards.yml)

I fact I don't know how to attribute a PR to a different one.

Thanks!

tarikgraba:
@trabucayre what are the fields Memory/Flash for in FPGA.yml?

trabucayre:
**memory** is if openFPGALoader can load bitstream into SRAM (ie volatile storage)
**flash** is if the bitstream may be wrtten into internal or external flash (ie non volatile storage)

For example **Zynq** has *NA* for **flash**, because the flash is accessible only by PS and **xfm** (a PROM device) has *NA* for **memory** because it's a JTAG flash.

For cycloneV SoC, theoritically it's possible to write bitstream into connected flash, but unfortunately I'm currently unable to do that.

trabucayre:
Thanks!
But the commit will be with your address instead of @sedwards-lab one

tarikgraba:
I have marked the flash as not tested and changed one of the commits

sedwards-lab:
No worries that the commit is under a different name. I'm just glad that support was added

trabucayre:
Applied to master!
Thanks!

Issue #47, opened 2020-09-03
on-the-fly decompression and spi-over-jtag


emard:
HI!

I suggest gzip on-the-fly decompression to upload bitstreams.
Especially it's good for spi-over-jtag bitstreams database
each one from 2MB becomes only 3.3KB with gzip -9

trabucayre:
Hi
It's not clear for me:
- You propose to store spiOverJtag bridge in gzip format, uncompress this file before loading in RAM ? or
- sending the gzip by spiOverJtag, uncompressing this one in the FPGA before writing the FLASH?

emard:
I propose to
support zlib on-the-fly decompression inside of openFPGALoader
and for flashing first search for spioverjtag<idcode>.bit.gz
on-the-fly decompress .gz using zlib support and program it and
then flash like normal.

It will significantly reduce spioverjtag bitstream size in the binary
installation of openFPGALoader

and I also propose to support reading bitstream from standard input,
so that user can do for example:

zcat bitstream.bit.gz | openFPGALoader

GbGp:
Xilinx FPGAs already supports a compressed bitstream format that can be loaded with openFPGAloader as usual.
To enable this feature it should be sufficient to just set ```set_property BITSTREAM.GENERAL.COMPRESS TRUE [current_design]``` constraint. Other than saving space this can make the flashing operation slightly faster.

emard:
HI

I know xilinx lattice and others support compressed bitstream,
but additional compression efficiency (file even shorter) can be
achieved by compressing that with gzip -9 and that was the
reason I was asking for this feature.

Instead of this, alternative would be to support reading bitstream
from unix pipe (stdin) so I can use for example

zcat bitstream.bit.gz | openFPGALoader

trabucayre:
I'm agree with both proposition.
A small test with xilinx compression and/or gzip shows:
| size  | compression |
|--------:|--------------------|
|2192129 | without xilinx compression |
|  236791 | with xilinx compression |
|     3390 | without xilinx compression + gzip -9 |
|   15910 | with xilinx compression + gzip -9 |

Clearly gzip applied on non compressed file is smaller, but with both the flashing time is faster + file small
The question is to determine which library to use to be compatible with linux, macOS and windows.

For unix pipe, my problem is a bit different: I use file extension to determine file type (it's diffcult or impossible to determine by content itself), with this approach it's required to find a way to select correct bitstream parser without extension information

emard:
[...]

PR #180, opened 2022-02-21
Add board and cable defaults for genesys2

jrrk2:
Added patches to allow you to just type:

openFPGALoader --board genesys2 picosoc.bit -f

Duplicated the bitstream for ffg900 as this seemed the easiest way. There still exists about a 30-second delay after powering up which seems much longer than the same operation with Vivado (just a curiosity at the moment).


unbtorsten:
Is there a particular reason as to why you did not use the default cable option in https://github.com/trabucayre/openFPGALoader/pull/180/files#diff-9854d3303f96a5d1c4a1bcabfaf4f3e8275c50755d8a02c1fabf2ce69dc3657bR165 ?

jrrk2:
No reason, except the target seems to like that frequency, and some of the other Digilent targets seem to specify the speed explicitly.

trabucayre:
It's true, default value with `CABLE_DEFAULT` is 6MHz so here `CABLE_MHZ(6)` is useless.
`CABLE_MHZ` is used when someone has tested with an highest frequency or when freq must be lowered because of a slow device.
Otherwise LGTM

hansemro:
Just added "xc7k325tffg900" to src/boards.hpp for kc705 and confirmed the same bitstream works.

```
openFPGALoader -b kc705 --verify -f led_sw.bit
write to flash
Open file DONE
Parse file DONE
use: /home/hansemro/vlsi/work/kintex7/openFPGALoader/built/share/openFPGALoader/spiOverJtag_xc7k325tffg900.bit.gz
load program
Done
Detected: micron N25Q128 256 sectors size: 128Mb
00000000 00000000 00000000 00
Done
Done
Verifying write (May take time)
Read flash : [==================================================] 100.00%
Done
```

unbtorsten:
> Just added "xc7k325tffg900" to src/boards.hpp for kc705 and confirmed the same bitstream works.

Did you commit any changes to the repo at this time? Please note that Xilinx' KC705 evaluation board uses the same package xc7k325tffg900 as the Digilent Genesys2 board at this time. However, people may recognize them as different boards.

hansemro:
I should clarify. I meant that minimal changes are needed on top of jrrk2's work to support kc705. @jrrk2 @trabucayre Is it okay to include kc705's support (by updating doc/boards.yml and src/boards.hpp) within this pull request?

unbtorsten:
@jrrk2 I have send you a pull request to fix the cable setting: https://github.com/jrrk2/openFPGALoader/pull/1
Assuming that pull request to be merged (this will automatically update the files for this pull request), I can confirm successfully loading a functional blinky onto the Genesys2.

trabucayre:
Could you just update `Makefile` xdc file and `build.py` before I merge?

For bitstream not working after power cycle: It makes me think to a default flash mode not compatible.
For alchity au flash must be configured to works on 4bits, fpga don't know and it's fails -> the bitstream must be conifgured for x1 to avoid boot failure.

trabucayre:
please add xdc file, build.py and Makefile update Thanks

jrrk2:
I'm sorry, I am not clear how build.py should be changed. I suppose we can copy ./spiOverJtag/constr_xc7k_ffg676.xdc to ./spiOverJtag/constr_xc7k_ffg900.xdc but I do not see how these are in any way device size independent. I think you are in the best position to decide whether it is OK to just copy bitstreams in the build process or whether they have to be built from scratch. As an unrepentant hacker I am the worst person to consult on matters of style.

jrrk2:
OK, if you think that's the most efficient way to do it. Until now I haven't come across pull requests to pull requests, I have to be honest it confuses me somewhat. Just as well I am due to retire in May!

unbtorsten:
> please add xdc file, build.py and Makefile update Thanks

@jrrk2 Please accept PR https://github.com/jrrk2/openFPGALoader/pull/2 to update this PR as requested.
To my surprise, writing to flash works with the bitstream, that you previously copied from ffg676 package. The pinout is entirely different. Anyhow, the bitstream can now be generated using the ffg900 xdc-file. I confirm that this yields functional output as well.

Loading from flash takes about 15s. I still think that it should be possible to expedite this, but that should be subject to a new issue. I consider this PR and the issue at https://github.com/kintex-chatter/xc7k325t-blinky-nextpnr/issues/14 to be resolved with this update.

trabucayre:
@unbtorsten Thanks to confirm the pinout by package for kintex follows same rules as all xilinx's devices (artix, spartan6, ...). I have the same question about current bitstream. If pinout is different spiOverJtag can't communicate with flash chip.

For loading delay it's strange. openFPGALoader do no bitstream's modifications, so the FPGA must take same time for the same bitstream... Maybe vivado do something like compression or change the mode (quad instead of single).
It's maybe interesting to dump flash after programing with vivado and compare to original one to see if something is different...

spiOverJtag/constr_xc7k_ffg900.xdc
```diff
@@ -0,0 +1,10 @@
+set_property CFGBVS VCCO [current_design]
+set_property CONFIG_VOLTAGE 3.3 [current_design]
+set_property BITSTREAM.CONFIG.SPI_BUSWIDTH {4} [current_design]
+
```
trabucayre, line 4: with adding:
`set_property BITSTREAM.GENERAL.COMPRESS TRUE [current_design]`
bitstream is smaller, so loading is faster

jrrk2:
I'm pretty sure Vivado uses the quad mode for this flash by default. Unfortunately I don't know how it gets it into quad mode, perhaps it is encoded in the flash header somehow.

spiOverJtag/constr_xc7k_ffg900.xdc
unbtorsten: Excellent point. Let me try that. As this does not change functionality of this PR, I would open a separate PR (and not go through jrrk2), if I see any benefit from this.

unbtorsten:
> I'm pretty sure Vivado uses the quad mode for this flash by default. Unfortunately I don't know how it gets it into quad mode, perhaps it is encoded in the flash header somehow.

Right... a comparison with a project of mine yields this:
```
set_property CONFIG_MODE SPIx4 [current_design]
```

I'll be adding both and get back to you once I know more.

jrrk2:
It should be surprise that the same bitstream works in ffg676 and ffg900 for the same device. After all, the only difference is that certain pins are not bonded out. Since the initialisation is fundamental to the hardware bootstrap those certainly will not be changed between different bond outs of the same device size.

spiOverJtag/constr_xc7k_ffg900.xdc
trabucayre: As mentioned in #47 I've done some test (but for artix) and with this option the bitstream size is divided by 10.

jrrk2:
The pins in the XDC file are just grid references to the ball grid array or pin grid array as appropriate. The mapping to internal pins will be done by an internal substrate (which is just a very fine resolution PCB). For sure the silicon is the same between these two, as is the prjxray extraction process.

trabucayre:
Maybe but it's seems a lucky try. Maybe it's not true for all devices / package. It make sense to have correct one to be sure to not have future side effects.

jrrk2:
If you don't believe my rationale you only need to run bit2fasm to see if there is any difference between the bitstreams. If there isn't it proves my theory, if there it might or might not be due to the random seed.

trabucayre:
It's really interesting thing yes! I have no kintex but I have to try with artix to see if it's the same. Maybe it's possible to reduce number of bitstream stored in repository thanks to that!

jrrk2:
I would not expect that rule to hold for the Virtex7 series that we have not got onto yet. As I understand it, these devices use multiple chips within the package with a 3-D clock distribution network. There would be plenty of scope for different possible JTAG chains in this case with the same silicon.

unbtorsten:
Bottom line, I think this PR is safe to merge at this time? Discussion of how to expedite boot time is moved to a dedicated issue at https://github.com/kintex-chatter/xc7k325t-blinky-nextpnr/issues/24

trabucayre:
Applied!

unbtorsten:
Great, thank you! That helps us simplify procedures over at the kintex-chatter corner.

trabucayre:
And thanks to you for contributions and informations!

PR #183, opened 2022-02-24
Darwin cmake config is missing Security framework

jrrk2:
If openFPGAloader is built on Darwin(OSX), it fails with undefined symbols under certain O/S versions, in particular OSX 12.2.1, AppleSilicon (the latest thing). This patch fixes the problem.


trabucayre:
I had planned to add this by myself. You are faster :)
Applied!
Thanks!

PR #184, opened 2022-02-24
Add CI for macOS

jeanthom:
This PR adds macOS builds in CI.


trabucayre:
Good idea to check macos too!
I don't know if it's possible to add artifact too -> fresh binary version up to date ?

Thanks

PR #186, opened 2022-02-27
Fix debian command

pepijndevos:
* add missing backslash
* use sudo like all other commands
* might as well use apt for nice progress meter and stuff


trabucayre:
Aplplied!
Thanks!

PR #190, opened 2022-03-11
spiOverJtag: compress the kintex7 bitstreams

rwhitby:
By adding a few properties to the constraints file, the resulting spiOverJtag bitstreams take far less time to load.

Tested on QMTech Kintex-7 core board and Digilent Genesys2 board.


trabucayre:
LGTM!
Applied
Thanks!

Issue #182, opened 2022-02-23
brew reinstall --build-from-source --verbose --debug  openFPGAloader not working on AppleSilicon


jrrk2:
brew reinstall --build-from-source --verbose --debug  openFPGAloader does not work on AppleSilicon under OSX 12.2.1

When installed as a binary, it does work  but the formula is outdated. The actual error is in libusb which installs without errors, but when linked with openFPGAloader, fails with the following errors:

[...]

PR #200, opened 2022-03-22
CMakeLists: fix general code styling, use dynamic linking on macOS

jeanthom:
* Making the indentation even across the whole file
* Use dynamic linking on macOS and fix its pkg-config usage in the process. This should render openFPGALoader agnostic to the location of Homebrew's prefix, and solve compatibility issues with MacPorts. It also removes all the dirty Make hacks which were necessary for macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,3 @@
-
 cmake_minimum_required(VERSION 3.0)
 
 # set the project name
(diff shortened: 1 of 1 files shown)


jeanthom:
This PR should fix #182

trabucayre:
LGTM.
If it simplify/improve the use with macOS  It's a really good thing!
Thanks!

PR #201, opened 2022-03-22
doc/guide: Add instructions for manual installation on macOS

jeanthom:
This PR adds instructions explaining how to install openFPGALoader on macOS from sources, using packages from Homebrew.


trabucayre:
Great!
Having to choice to use binary package (when available) or to install manually is really useful.
Thanks!

PR #203, opened 2022-03-24
Spartan3 (xc3s500evq100)

rstephan:
This adds support for the Xilinx Spartan3E  (xc3s500evq100) and a Microchip SST25VF040B flash chip.
They are found, for example, on a Papilio One FPGA board.
 
The spiOverJtag gets a bit messy with all the `ifdef`, but I found no other way to keep it in one file.
Sadly the flash only supports 1 byte per Byte-Program (0x02) instruction. For now there is no time for me to implement the AAI-Word-Program feature. It's really slow ~4 minutes or so!

The `program_mem()` function also needed a bit for tweaking. The algorithm came from a svf-file, I haven't found the documentation on Xilinx website.


trabucayre:
Thanks!
Yes spiOverJtag is a bit complex: maybe I have to think to split the file into more.
Is it possible to have two PR: one for spartan3 and one for flash ?
doc must be updated to reference this device and the board.
I have to check for addition in `program_mem` (bsd files for xc3s are in `ISE` folder (Xilinx/14.7/ISE_DS/ISE/spartan3/data))

trabucayre:
For `xilinx.cpp` update: I have read */somewhere/Xilinx/14.7/ISE_DS/ISE/spartan3e/data/xc3s500e_vq100_1532.bsd*. This file provides the full procedure to load a bistream. It's true it's a bit different (and in fact more simple than current `program_mem`).
I hesitate to add specific a method (as done for *xcf*, *xc9500xl* and *xc2c*) and to use it instead of generic one. The drawback is to add a new method adding complexity for maintainability

trabucayre:
Ultimately, I have added method to load bistream for xc3s (it's more or less same as `program_mem` but with delay more stricts and some tricks from xx_1532.bsd files).
If you could test that.
Thanks

rstephan:
I will try your code, no problem. But give me some time.

trabucayre:
PR #204 is now applied.
For "Wait Until Done" it's weird: with my xc3s200 it's works... Have you a reference for a board with xc3s500e (or similar) ?
I use `shiftIR` with `BYPASS` to have access to the *Instruction Capture Values* (UG332 v1.7 p.208 table 9.6). Theoritically `shiftDR` is not required...
A temporary work around is to bypass this check when *spartan3e* devices

trabucayre:
I have searched for both board with no success: Out of stock...
I have to try with my OLS: it has an spartan3E too.
But it's weird: bsd files didn't shows real differences about programming with E and non-E devices.

rstephan:
Maybe not the right place for a "commercial", but my BASYS2 is from a german ebay auction. There are still 2 boards left. I don't know the shipping details for your country, see yourself.

trabucayre:
It's looks good: if it's possible to have access to this hardware there is no wrong place.

trabucayre:
My OLS does the job: it's true done bit goes never high. I have adapted behaviour according to datasheet.
Now load to RAM finish correctly

rstephan:
Thanks for your help.

trabucayre:
Happy to help!
It's maybe time to close this PR ?
Feel free to open a new PR or issue for improvement / regression.
Thanks!

rstephan:
Right!

PR #204, opened 2022-03-25
Support for the Microchip SST25VF040B flash

rstephan:
Programming is slow, only one byte at a time :-(

See https://github.com/trabucayre/openFPGALoader/pull/203

### DIFF
--- a/src/spiFlash.cpp
+++ b/src/spiFlash.cpp
@@ -311,10 +311,6 @@ int SPIFlash::erase_and_prog(int base_addr, uint8_t *data, int len)
 		if ((status & 0x1c) != 0)
 			must_relock = true;
 	}
-	if ((_jedec_id >> 8) == 0xbf258d) {
-		if ((status & 0x1c) != 0)
-			must_relock = true;
-	}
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM.
Yes AAI mode seems mandatory to reduce write time!
Just two thing: 
- there is not TB so this is a use case to integrate
- BP bits can't be updated if `BPL` (status register bit 7) is high -> again a use case to integrate into `enable_protection` and `disable_protection`

It's not to ask you to add these feature, I can, of course, add those by myself once this PR is applied.

Thanks

rstephan:
At least the TB settings are removed form my side.
I am fine, if you can implement `enable_protection`/`disable_protection` yourself.
You know better how to find the right place in your code. ;-)
Thanks!

trabucayre:
Could you just remove test about bp lines 314-317.
For bypassing TB, I will do that by myself, it's okay.
Thanks

trabucayre:
Applied thanks!
I have added logic to handle flash without TB bit.

PR #205, opened 2022-03-25
Spartan3E (xc3s500evq100) and Papilio One

rstephan:
Support for:

- JTAG over SPI
- Xilinx xc3s500evq100
- Papilio One board

I have added a short documentation and left out the `program_mem()` for now.
Give me your feedback for how the add or improve the programming function.
Thanks.

This is the second half, see https://github.com/trabucayre/openFPGALoader/pull/203


src/part.hpp
```diff
@@ -31,6 +31,7 @@ static std::map <int, fpga_model> fpga_list = {
 	{0x03651093, {"xilinx", "kintex7", "xc7k325t", 6}},
 
 	{0x01414093, {"xilinx", "spartan3", "xc3s200",  6}},
+	{0x41c22093, {"xilinx", "spartan3", "xc3s500e", 6}},
```
trabucayre, line 34: When it's possible (no ambiguities between FPGAs) high nibble (revision) is set to 0 -> it's not required to add more than one entry if an FPGA has more than one revision.

PR #208, opened 2022-03-28
part: Add kintex7 variant xc7k410t

skooNI:
Hi!

This PR adds support for the xc7k410t variant of the kintex 7. This is the fpga part used in the USRP X310. I've tested flashing the SRAM on this device using the integrated Digilent JTAG cable. I've also tested the existing xc7k325t support with a USRP X300 and verified flashing the SRAM works as well.

Please let me know if y'all have any feedback. Thanks for such a cool project!

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -177,6 +177,7 @@ Xilinx:
     Model:
       - xc7k160t
       - xc7k325t
+      - xc7k410t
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM.
Applied thanks!
Feel free to test flash write for both USRP (spiOverJtag support for xc7k325t is already done, xc7k410t has to be added)
And of course, you can add both X3x0 in the supported board list.

PR #218, opened 2022-04-29
efinix: support loading a bitstream to SRAM using only JTAG pins (no GPIO)

ejdan:
Tested with Trion T120 BGA324 Development Kit but should work on all T20BGA324, T20BGA400, T35, T55, T85, and T120. This makes it possible to use generic FTDI JTAG cables which do not have the extra reset, oe and cs pins connected.


trabucayre:
LGTM.
Applied thanks!

Issue #221, opened 2022-05-11
Add skip args for SRAM + Flash operations


el-coder-sb:
For debugging purposes it would be great to have cmd line args to skip one part of both SRAM and Flash operation in case of e.g. Altera.

el-coder-sb:
Done in #222

PR #222, opened 2022-05-11
Add skip sram cmd

el-coder-sb:
Ref #221

Just a first quick approach (implemented only for Altera). Do you think this is the right direction?
Beside the added cmd line args a significant change is moving prepare_flash_access().

Other possbile ways I can see would be to add different ops to prog_type_t.

### DIFF
--- a/src/spiInterface.cpp
+++ b/src/spiInterface.cpp
@@ -11,13 +11,14 @@
 #include "spiFlash.hpp"
 
 SPIInterface::SPIInterface():_spif_verbose(0), _spif_rd_burst(0),
-	_spif_verify(false)
+	_spif_verify(false), _skip_load_bridge(false)
 {}
 
 SPIInterface::SPIInterface(const std::string &filename, uint8_t verbose,
-		uint32_t rd_burst, bool verify):
+		uint32_t rd_burst, bool verify, bool skip_load_bridge):
 	_spif_verbose(verbose), _spif_rd_burst(rd_burst),
-	_spif_verify(verify), _spif_filename(filename)
+	_spif_verify(verify), _skip_load_bridge(skip_load_bridge),
+	_spif_filename(filename)
 {}
 
 /* spiFlash generic acces */
--- a/src/spiInterface.hpp
+++ b/src/spiInterface.hpp
@@ -20,7 +20,7 @@ class SPIInterface {
  public:
 	SPIInterface();
 	SPIInterface(const std::string &filename, uint8_t verbose,
-			uint32_t rd_burst, bool verify);
+			uint32_t rd_burst, bool verify, bool skip_load_bridge = false);
 	virtual ~SPIInterface() {}
 
 	bool protect_flash(uint32_t len);
@@ -97,6 +97,7 @@ class SPIInterface {
 	uint8_t _spif_verbose;
 	uint32_t _spif_rd_burst;
 	bool _spif_verify;
+	bool _skip_load_bridge;
  private:
 	std::string _spif_filename;
 
--- a/src/altera.cpp
+++ b/src/altera.cpp
@@ -25,9 +25,10 @@
 
 Altera::Altera(Jtag *jtag, const std::string &filename,
 	const std::string &file_type, Device::prog_type_t prg_type,
-	const std::string &device_package, bool verify, int8_t verbose):
+	const std::string &device_package, bool verify, int8_t verbose,
+	bool skip_load_bridge):
 	Device(jtag, filename, file_type, verify, verbose),
-	SPIInterface(filename, verbose, 256, verify),
+	SPIInterface(filename, verbose, 256, verify, skip_load_bridge),
 	_svf(_jtag, _verbose), _device_package(device_package),
 	_vir_addr(0x1000), _vir_length(14)
 {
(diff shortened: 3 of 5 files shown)


trabucayre:
It's a good idea but I think a bit too complex:
- spiOverlay is loaded in `prepare_flash_access` for both *xilinx* and *altera* class
- reset command is send in `post_flash_access` for both *xilinx* and *altera* class
So It make more sense to add these two options at constructor level for both classes and in dedicated methods skip content. With this approach only *altera* and *xilinx* classes are impacted in a first time (bypassing reset may be extended to all device class).
I know internal details are not described at all :-/
For params `skip-load-bridge` or `skip-load-spioverjtag` is more clear, and `detect-flash` instead of skip-flash.
Thanks!

el-coder-sb:
Thanks for your feedback!

> So It make more sense to add these two options at constructor level for both classes

Yepp, of course, makes sense (working too rarely with c++/classes).

What I have in mind is to seperate the steps `prepare_flash_access`, `flash.read_status_reg`, `flash.erase_and_prog`, `flash.verify` and `post_flash_access` in such a way that I can do any number of these five steps according to cmd line arguments. That would, IMHO, make debugging much easier for the future. Do you agree that it is worth to target?

Now, how to get to that point? I added some flags to do for some of the five steps (and moved and renamed the ones I proposed before). See [354b37e](https://github.com/trabucayre/openFPGALoader/pull/222/commits/354b37ec68e58ca10aed7dea8e9f8c758e1939a6). However `SPIInterface::write` looks somewhat ugly as the flags are checked in different functions (Altera::load_bridge and SPIInterface::write).

trabucayre:
This look better! Thanks!
Just again some remarks:
- `spiInterface` is a mother class for most of device driver, it's make sense to store params at this level instead of device level (this reduce again the amount of modifications)
- `skip-flash-access` seems a bit useless: you could load bridge manually
- `skip-flash-write` again seems not really useful I think a detect method used at main level do make sense (I'm working on)

In my mind a `skip-reset` needs to be added to avoid resetting device after playing with the flash
`detect` param behaviour may be updated:
- if used with SRAM: list of FPGAs is displayed
- if used with FLASH: it must call specific method to display flash informations.

In fact I'm try to keep a reasonable number of args: I try to find compromise between too many and usefull. Parsing args take time and a too big list is difficult to maintain and not easy, for user, to understand.

el-coder-sb:
> spiInterface is a mother class for most of device driver, it's make sense to store params at this level instead of device level (this reduce again the amount of modifications)

Done, OK this way? (or did I misunderstand you?)

> skip-reset needs to be added

Not sure where to implement this. May be we let it for another pull-request? Should it be at device level (i.e. inside `post_flash_access`) or at SPI-level?

src/spiInterface.cpp
```diff
@@ -11,13 +11,21 @@
...
-	_spif_verify(verify), _spif_filename(filename)
+	_spif_verify(verify), _skip_load_bridge(false),
+	_spif_filename(filename)
+{}
+
+SPIInterface::SPIInterface(const std::string &filename, uint8_t verbose,
+		uint32_t rd_burst, bool verify, bool skip_load_bridge):
+	_spif_verbose(verbose), _spif_rd_burst(rd_burst),
+	_spif_verify(verify), _skip_load_bridge(skip_load_bridge),
+	_spif_filename(filename)
```
trabucayre, line 28: Why adding a new constructor?
You can simply add `skip_load_bridge` in the existing one and provides a default value in .hpp declaration

src/spiInterface.hpp
```diff
@@ -21,6 +21,8 @@ class SPIInterface {
 	SPIInterface();
 	SPIInterface(const std::string &filename, uint8_t verbose,
 			uint32_t rd_burst, bool verify);
```
trabucayre, line 23: here add `bool skip_load_bridge = false`

trabucayre:
LGTM.
Yep for reset I think it's the same approach, at device level in post_flash. It's again only relevant for a subset of devices and it's globally better to let device device driver decide how to deal with this option.

src/spiInterface.hpp
el-coder-sb: Thanks for guiding me. I wasn´t aware that cpp supports this kind of default value setting. I have never really  written cpp code.

el-coder-sb:
I think now it looks good.

trabucayre:
LGTM!
I have a last request: could you rewrite your branch to reduce list of commits. It's to keep history as clean as possible.
If you can't maybe I could apply your patch manually.
Thanks

el-coder-sb:
This way? (see action force-push above). I´m not very familiar with rewriting git history, so I simply squashed everything. Is that OK for you? For sure, two commits would be some more reasonable (one for SPI interface + args and one for Altera) but I would have to look for such (rewriting) action.

trabucayre:
Yep it's good to me.
Usually I update a PR like that:
1. modify source
2. `git add xxx`
3. `git commit --amend` (left content unchanged)
4. `git push --force

This way I'm able to rewrite already pushed file/commit without having to modify history. After that if I have others commits I use classic way.

trabucayre:
Applied!
Thanks!

trabucayre:
Added skip-reset too

PR #223, opened 2022-05-12
--flash-sector CFG0 needed for at least my MachXO3D

bluesceada:
Needed for at least the MachXO3D Breakout Board. But from reading the source code, there is no default value for flash-sector, so it would always fail, if it is not given. Unfortunately the .bit file seems to can not be flashed at all on that board based on command line options.


trabucayre:
Before merging this PR I need to check this device.
It a bit strange CFG0 is only required when bitstream (.jed) is written into flash, not when the bitstream (.bit) is loaded in SRAM.
In fact support for machXO3D has been done by @ghost

I have ordered some month ago amachXO3D eval board but not already used.
Thanks!

bluesceada:
I can not flash the bit file at all on my board. Not sure what I am supposed to do to be able to do it. The official vendor software doesn't have options like CFG0 for writing to SRAM, but I also tried to do it, without success. Maybe I should rather open an issue?

trabucayre:
CFG0 is related to a specific internal flash area: for SRAM this doesn't make sense.
With vendor tool are you able to load your .bit into the SRAM?
For internal flash you have to use .jed file

bluesceada:
On the MachXO3D Breakout Board, in Lattice Diamond, you can program the .jed file to CFG0 using the flash program mode, and the .bit file (without any additional options) using the SRAM program mode.

trabucayre:
For `.jed` file: the command `openFPGALoader --flash-sector CFG0 top.jed` is working.
For `.bit` file I see two issues:
- an error is raised because of missing flash-sector option (this one is only required for flash mode): I push the fix ASAP
-  `bit` preamble in `laticeBitParser` is not correctly parsed: Error message is raised here too and consecuently nothing is done... I search a fix but I have to verify fix doesn't add regression for others devices (`ECP5`, `machXO2`, `machXO3`, `nexus`,...)

trabucayre:
I have pushed both fixes. Could you retry with the master branch?
Thanks!

trabucayre:
It's good to you?

bluesceada:
Sorry for the late reply, and thanks a lot! Writing the SRAM now works almost in an instant, while Lattice Diamond takes probably around a minute for it, that's great! I verified that it actually works. Flashing also is very fast, it needs the `--flash-sector CFG0` as I wrote above. Thanks again!

trabucayre:
Great news! Thanks!

Yes, lattice programmer is awfully slow... And it's true for all devices... But when reading for USB transactions it seems a lot of garbage (maybe to obfuscate stream...).

Your PR must be adapted/redone to add specifics details for machXO3D when user want to write into flash.

bluesceada:
I updated the patch-1 branch to match to the current master and changed it such that it only mentions CFG0 for the MachXO3D Breakout Board. Do you want me to add the board to the ./doc/boards.yml file in some way? I see the MachXO3D Evaluation Board was added. However, the one I have here and which works for me is the MachXO3D Breakout Board ( https://www.latticesemi.com/products/developmentboardsandkits/machxo3d_breakout_board ).

trabucayre:
Looks good.
Yes, if it's possible to add this breakout board too it's a good thing!
Just a small request: could you rewrite branch history. It's to keep after applied a , as much as possible, clean history.

Thanks

trabucayre:
LGTM.
Applied!
Thanks!

PR #225, opened 2022-05-14
spiOverJtag: support xc7k160tffg676

hansfbaier:
Just added this Kintex chip, while I am at it.
Flashing seems to work,
but after booting the bitstream does run.
Initially it didn't but that seemed to be because of wrong dip switch settings on the board.
When I boot the flash with the vendor utility the bitstream worked fine.
So, all good as it seems!

```
$ openFPGALoader -c ft2232 --fpga-part xc7k160tffg676 -f ./xc7k160t.runs/impl_1/blinky.bit
write to flash
Open file DONE
Parse file DONE
use: /usr/local/share/openFPGALoader/spiOverJtag_xc7k160tffg676.bit.gz
load program
Done
Detail: 
Jedec ID          : 01
memory type       : 02
memory capacity   : 20
EDID + CFD length : 4d
EDID              : 8000
CFD               : 
flash chip unknown: use basic protection detection
Done
Done
```


trabucayre:
Applied!
Thanks!
flash chip is unknown -> I have added the model.

hansfbaier:
Great, thanks!

PR #226, opened 2022-05-14
Add initial support for ZCU106 development board

MPLew-is:
This works as expected for a bare-bones "hello world" bitstream (tying LEDs to switches) on my ZCU106 board.


trabucayre:
Applied!
Thanks!

PR #227, opened 2022-05-14
fix(part.hpp): use uint32_t instead of int for fpga_list index

Xiretza:
This resulted in an error with GCC 12:

```
In file included from src/main.cpp:29:
src/part.hpp:148:1: error: narrowing conversion of ‘2165379139’ from ‘unsigned int’ to ‘int’ [-Wnarrowing]
  148 | };
      | ^
```

### DIFF
--- a/src/part.hpp
+++ b/src/part.hpp
@@ -6,6 +6,7 @@
 #ifndef PART_HPP
 #define PART_HPP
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Good catch! Thanks!
Applied!

PR #228, opened 2022-05-16
Correct part number for ZCU106 board

MPLew-is:
Not sure if this matters at all, but it looks like I made a copy-paste error in #226 - this is the correct part for that board.


trabucayre:
Applied thanks!
This information is only relevant for board where it's possible to write the bitstream into flash: here it's not true, but it's always better to have correct code ;-)

PR #236, opened 2022-06-16
Add GW2A support

Icenowy:
Checksum implementation is TODO.


trabucayre:
LGTM.
Applied to master.
Thanks

trabucayre:
One way to compare checksum read and bitstream checksum is simply to use one provided by the bitstream (in a first time).
Second thing is about external flash access: current implementation is the bitbang approach, there is another way but could only be used when FPGA is configured to read bitstream from external flash.

=> ASAP tang primer 20k will be available I have to order one to try to fix checksum and implement second (fastest) external flash access.

PR #239, opened 2022-07-02
cmsisDAP: better error message if support is not enabled

BBBSnowball:
before:

```
$ ./openFPGALoader -c cmsisdap /dev/null
Jtag: unknown cable type
JTAG init failed with: std::exception
```

after:

```
$ ./openFPGALoader -c cmsisdap /dev/null
Jtag: support for cmsisdap was not enabled at compile time
JTAG init failed with: std::exception
```

It took me a while to realize that the error was not due to a typo or misunderstanding. NixOS is building it without hidapi so support for cmsisdap is automatically disabled.


trabucayre:
Good catch!
It's true a better message for cable not supported is definitively a good approach!
Applied
Thanks!

PR #242, opened 2022-07-18
Add JTAG driver for libgpiod bitbanging

niklasekstrom:
A driver is added that allows bitbanging JTAG on a Linux host with GPIO pins, such as a raspberry pi.


CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ else()
 	option(ENABLE_UDEV "use udev to search JTAG adapter from /dev/xx" ON)
 endif()
 option(ENABLE_CMSISDAP "enable cmsis DAP interface (requires hidapi)" ON)
+option(ENABLE_LIBGPIOD "enable libgpiod bitbang driver (requires libgpiod)" ON)
```
trabucayre, line 14: Maybe here or below a test about OS. I'm not familiar with macOS and windows and not sure `/dev/gpiochipxx` is something allowed.

src/cable.hpp
```diff
@@ -69,6 +70,7 @@ static std::map <std::string, cable_t> cable_list = {
 	{"usb-blaster",        {MODE_USBBLASTER,   {0x09Fb, 0x6001, 0,           0,    0,    0,    0,    0}}},
 	{"usb-blasterII",      {MODE_USBBLASTER,   {0x09Fb, 0x6810, 0,           0,    0,    0,    0,    0}}},
 	{"xvc-client",         {MODE_XVC_CLIENT,   {}}},
+	{"libgpiod",           {MODE_LIBGPIOD_BITBANG, {}}},
```
trabucayre, line 73: `#ifdef ENABLE_LIBGPIOD` to avoid to display this cable if support is disabled?

src/jtag.cpp
```diff
@@ -130,6 +133,13 @@ void Jtag::init_internal(cable_t &cable, const string &dev, const string &serial
 #else
 		std::cerr << "Jtag: support for xvc-client was not enabled at compile time" << std::endl;
 		throw std::exception();
+#endif
+#ifdef ENABLE_LIBGPIOD
+	case MODE_LIBGPIOD_BITBANG:
+		if (pin_conf == NULL)
+			throw std::exception();
```
trabucayre, line 140: An error message to explain why there is a failure?
And as for **xvc** and **cmsis** an `#else` to display a message when user try to use this driver if it's disabled.

src/libgpiodJtagBitbang.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+		for (int j = i + 1; j < sizeof(pins) / sizeof(pins[0]); j++) {
+			if (pins[i] == pins[j]) {
+				display("Two or more pins are assigned to the same pin number %d\n", pins[i]);
+				throw std::runtime_error("Two or more pins are assigned to the same pin number\n");
+			}
+		}
+	}
+
+	_chip = gpiod_chip_open_by_name(chip_dev.substr(5).c_str());
```
trabucayre, line 72: `gpio_chip_open` takes a `/dev/gpiochipxx` it's avoid to cut `/dev/`

src/libgpiodJtagBitbang.cpp
```diff
@@ -0,0 +1,225 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+ * Copyright (C) 2020 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
```
trabucayre, line 3: Wrong copyright :)

src/libgpiodJtagBitbang.hpp
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+ * Copyright (C) 2020 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
```
trabucayre, line 3: ditto

src/libgpiodJtagBitbang.cpp
niklasekstrom: In what way is the copyright wrong? I based the code on the file ftdiJtagBitbang.cpp, and then added a line under your name that says that the libgpiod code was added by me.

src/jtag.cpp
niklasekstrom: That code is copy-pasted from the MODE_FTDI_BITBANG case above in the switch statement.

It looks like the `if (pin_conf == NULL)` check is not needed as pin_conf will always be non-NULL, seen by looking at the call site, so I'll just remove the check.

I'll add the `#else` message you suggested.

src/libgpiodJtagBitbang.cpp
niklasekstrom: Done.

src/cable.hpp
niklasekstrom: Will do.

src/libgpiodJtagBitbang.cpp
trabucayre: I'm not really aware about copyright rule (I must must learn more about).
This class is based on another but with modifications so it seems made sense to add you in copyright are adding a note like "based on xxx". You have done the job to adapts code for libgpiod.
Or simply adding a second copyright line :)

trabucayre:
LGTM.

trabucayre:
Applied to master
thanks!

PR #243, opened 2022-07-20
Gowin GW1NSR-4C MCU support

balika011:


src/gowin.cpp
```diff
@@ -153,6 +154,30 @@ Gowin::Gowin(Jtag *jtag, const string filename, const string &file_type,
...
+		fseek(fd, 0, SEEK_END);
+		int file_size = ftell(fd);
+		fseek(fd, 0, SEEK_SET);
+
+		mcufwdata.resize(file_size);
+
+		int ret = fread((char *) mcufwdata.data(), sizeof(char), file_size, fd);
+		fclose(fd);
+		if (ret != file_size)
+			throw std::runtime_error("Error: fail to read " + mcufw);
```
trabucayre, line 176: Maybe using RawParser is better for coherency

src/main.cpp
```diff
@@ -537,6 +539,13 @@ int main(int argc, char **argv)
 		}
 	}
 
+	if (args.conmcu == true) {
+		fpga->connectJtagToMCU();
+		delete(fpga);
+		delete(jtag);
+		return EXIT_SUCCESS;
```
trabucayre, line 546: Why quit here? It's done later

paller:
> Looks good but I hesitate between a specific method/call to program mcu or to consider it's gowin specific and consequently it's internal to the gowin class (if mcufw is not null then flash MCU otherwise flash bitstream).

I would opt for a specific/generic method to program the MCU in combined devices as GoWin is not alone in making MCU+FPGA devices. There is also the EOS S3 from QuickLogic.

src/gowin.cpp
balika011: Sure, i'm not really familiar the codebase, I was happy that it works.

balika011:
> if mcufw is not null then flash MCU otherwise flash bitstream

That won't really work, because the data (firmware/bitstream) for fpga and mcu are stored on the same flash. We need to erase before flashing and I don't know how to do partial erase over jtag.

src/main.cpp
balika011: True, that was a harm-less mistake.

trabucayre:
@paller you remark is good: EOS-S3, zynq and cycloneVSoc have a CPU into package. Currently only FPGA is supported but why not a minimal CPU support to load firmware without having to use two tools. It's something I need to think about.

@balika011 with Gowin I'm not sure if it's possible to erase only a specific area, so it may make sense to left this as it. So please just fix two remarks. Thanks

balika011:
> @balika011 with Gowin I'm not sure if it's possible to erase only a specific area, so it may make sense to left this as it. So please just fix two remarks. Thanks

From verilog it is.  GOWIN_EMPU internally connects the MCU to FLASH256K. Same thing that you can access under GW_USER_FLASH_256K.

trabucayre:
> > @balika011 with Gowin I'm not sure if it's possible to erase only a specific area, so it may make sense to left this as it. So please just fix two remarks. Thanks
> 
> From verilog it is. GOWIN_EMPU internally connects the MCU to FLASH256K. Same thing that you can access under GW_USER_FLASH_256K.

Maybe adding a note about that in doc/vendors/gowin.md may help users ?

trabucayre:
Applied to master.
Thanks!

PR #247, opened 2022-08-18
Add Xilinx Kintex UltraScale KCU035 IDCode

Novakov:
Tested loading bitstream with Trenz KCU035 SoM. After adding IDCODE it worked like a charm :)


trabucayre:
Applied Thanks!
Have you tried spiOverJtag?

Novakov:
No, only load to memory over JTAG. Actually I needed only recording of JTAG lines during programming and openFPGALoader was so much more useful than XIlinx tools :)

Issue #259, opened 2022-10-13
GW1N-9C: flash CRC error


jeanbie0:
Hello,
I've CRC error when I try to write the flash of the FPGA.

I succeed to write the SRAM without any problem and the code is running.

```
openFPGALoader --unprotect-flash -f -v -c libgpiod --pins 65:28:20:19 ./code.fs

write to flash
found 1 devices
index 0:
        idcode 0x100481b
        manufacturer Gowin
        family GW1N
        model  GW1N(R)-9C
        irlength 8
File type : fs
Parse file Parse ./code.fs:
checksum 0xb8e4
Done
DONE
bitstream header infos
CRCCheck: ON
Compress: OFF
ConfDataLength: 712
ProgramDoneBypass: OFF
SPIAddr: 00000000
SecurityBit: ON
idcode: 1100481b
loading_rate: 0
pollFlag: 350a2
erase SRAM pollFlag: 35082
pollFlag: 35082
pollFlag: 35082
pollFlag: 310a2
Done
pollFlag: 31422
pollFlag: 310a2
erase Flash Done
pollFlag: 11422
write Flash: [==================================================] 100.00%
Done
pollFlag: 11422
CRC check : FAIL
b8e4 0000
displayReadReg 00015421
        CRC Error
        Memory Erase
        Non-jtag is active
        Gowin VLD
        Security Final
        POR
```

Thank you

trabucayre:
Everything looks good the `--unprotect-flash` is not required here because is related to external flash.
But it's weird, last time I have seen CRC equal to 0 is with a too short sleep before reading (ie bitstream not completly loaded).
But with libgpiod (slow) I'm not sure if it's the reason. This device (seems the same as sipeed tang nano 9k) has been validated by many people.
Is it a commercial board (ie easy to order) and it's possible to you to test with a jtag cable ?
Thanks

trabucayre:
It's interesting: I always take into account max freq because it's not possible to work with an higher rate, but not min freq.
I have to do tests to see if this requirement must __absolutely__ be true: at least to print a warning when not met.

jeanbie0:
It was this.

I create a new cables based on libgpiod cable. This cable "jetson-nano-gpio" is only usable with Jetson Nano (Tegra X1) because of the structure of internal memory. but it can be easily adapted for other CPU.

With this I can have a clock frequency of 1.2MHz. Still too slow for the datasheet but the writing in flash is working.

[...]

PR #260, opened 2022-10-14
Add cable jetson-nano-gpio

jeanbie0:
Related to the issue #259.

Tested on Jetson Nano custom board with Gowin GW1N-9C.
Added this because this FPGA need a minimal clock frequency that can be reached with libgpiod.

`-DENABLE_JETSONNANOGPIO=ON` to enable.


trabucayre:
I have some questions:
- the most part of this new class is similar to `libgpiodJtagBitbang`: what do you think to inherint one of them from the other? And overriding only relevant methods?
- is this code is portable? It's seems to bypass **libgpiod** maybe it's possible to have something more generic -> possible to be used with others CPU?

I assume the **patch** file was added by error.

Thanks!

jeanbie0:
Hello,

Yes the class is similar. But I don't thing that one of them inherent from the other. They are more sister class and inherent from class `JtagInterface `. So it seems good like this.

No code is not portable for other CPU. It depends of the internal structure of the CPU's GPIO. (Address of ports, size of ports, structure of ports,...)

Yes patch file is an error.

trabucayre:
> Yes the class is similar. But I don't thing that one of them inherent from the other. They are more sister class and inherent from class `JtagInterface `. So it seems good like this.

I have do a diff between these two class: it's differs by CTOR, DTOR, `update_pins` and `read_tdo`. Rest of the logic remain the same. It's maybe possible to override corresponding `libgpiod`'s methods to keep logic from this one and simply add the logic required to a direct communication with CPU's controler?

> No code is not portable for other CPU. It depends of the internal structure of the CPU's GPIO. (Address of ports, size of ports, structure of ports,...)

It's make sense yes.

jeanbie0:
So I think the most clean will be to create a mother class `GPIOBitbang` from which `libgpiodBitbang` and `jetsonNanoBitbang` inherit. So in the future it will be easy to create a class for another CPU if necessary.

trabucayre:
Adding a super class seems the best way: it's maybe possible to use this one for ftdi bitbanging too.

trabucayre:
Applied.
I will see howto introduce this super class and if it's possible to adapt ftdijtagbitbang to this new structure.
Thanks!

PR #261, opened 2022-10-16
Limit automatic read burst length to 64K

petr-nechaev:
Reading of very large binaries (e.g. 32 MB) can cause stack overflow.
Put a reasonable limit on read burst length.

### DIFF
--- a/src/spiFlash.cpp
+++ b/src/spiFlash.cpp
@@ -382,8 +382,11 @@ int SPIFlash::erase_and_prog(int base_addr, uint8_t *data, int len)
 bool SPIFlash::verify(const int &base_addr, const uint8_t *data,
 		const int &len, int rd_burst)
 {
-	if (rd_burst == 0)
+	if (rd_burst == 0) {
 		rd_burst = len;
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM! Thanks!
Just a question: which FPGA you use? Most of FPGA's class with an external flash have a limit around 256B?
This high limit looks good!

petr-nechaev:
I boot Linux on Colorlight 5A-75B rev 8.0 with Lattice ECP5 LFE5U-25F-6BG256C see https://github.com/q3k/chubby75/blob/master/5a-75b/hardware_V8.0.md

Standard 32Mbit / 4MB SPI Flash was not enough to store the bitstream, OpenSBI and Linux Kernel so I had to change to 8MB. Actually I plan to use 64MB flash in the future for all my applications.

trabucayre:
OK. I have seen this limitation with my v7.0 too.
By the past, read burst for lattice was limited but it's seems I have lost this information during one spiInterface rewrite.
Applied thanks!

PR #263, opened 2022-10-22
msys2: Fix windows paths

### LINKED ISSUE #262: Xilinx flash programming fails using release
## Summary
After installing openFPGALoader package in MSYS2, loading a bitstream into a xillinx's config SRAM works, but loading into FLASH fails.

```
$ openFPGALoader.exe -b qmtechKintex7 -c tigard -f fpga/fpga.bit
write to flash
Board qmtechKintex7 has not default cable
found 1 devices

...

use: D:/a/_temp/msys64/mingw64/share/openFPGALoader/spiOverJtag_xc7k325tffg676.bit.gz
Error: fail to open D:/a/_temp/msys64/mingw64/share/openFPGALoader/spiOverJtag_xc7k325tffg676.bit.gz
Error: Failed to program FPGA: Error: fail to open D:/a/_temp/msys64/mingw64/share/openFPGALoader/sp
```
Note the file does exist, and has been installed.. But is located at `/msys64/mingw64/share/openFPGALoader/spiOverJtag_xc7k325tffg676.bit.gz`

## Detailed issue

The extra `D:/a/_temp/` prefix on the directory seems to originate from a hard coded path in the source code. `DATA_DIR` in a prefix for the bitfile.

https://github.com/trabucayre/openFPGALoader/blob/6a00466137ca8c20b519c7a5193b338f7e246935/src/xilinx.cpp#L315-L341

`DATA_DIR` is defined by CMAKE

https://github.com/trabucayre/openFPGALoader/blob/6a00466137ca8c20b519c7a5193b338f7e246935/CMakeLists.txt#L35

On the CI builds that end up as releases this value for MSYS2 builds becomes `D:/a/_temp/msys64/mingw64/share/`

gregdavill:
Fixes #262 

Following the example from msys2 (https://www.msys2.org/wiki/Porting/) "Hard-coded paths and relocation"
> Sometimes these paths are baked into shell scripts or pkg-config's .pc files. In that case, you should use sed in the PKGBUILD package() function to correct this back to the msys2 version of the path. 

The package step can strip the `D:/a/msys64/` or equivalent prefix that's added at compile time.

```
 -- Version installed from msys2 packages
$ strings /mingw64/bin/openFPGALoader.exe | grep mingw64/share
D:/a/msys64/mingw64/share/openFPGALoader/spiOverJtag_

 -- Version installed from this PRs CI build
$ strings /c/Users/greg/Downloads/openFPGALoader | grep mingw64/share
/mingw64/share/openFPGALoader/spiOverJtag_
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,7 @@ set(OPENFPGALOADER_SOURCE
 	src/jlink.cpp
 	src/lattice.cpp
 	src/progressBar.cpp
+	src/pathHelper.cpp
 	src/fsparser.cpp
 	src/mcsParser.cpp
 	src/ftdispi.cpp
@@ -138,6 +139,7 @@ set(OPENFPGALOADER_HEADERS
 	src/ice40.hpp
 	src/ihexParser.hpp
 	src/progressBar.hpp
+	src/pathHelper.hpp
 	src/rawParser.hpp
 	src/usbBlaster.hpp
 	src/bitparser.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,6 @@ set(OPENFPGALOADER_SOURCE
 	src/jlink.cpp
 	src/lattice.cpp
 	src/progressBar.cpp
-	src/pathHelper.cpp
 	src/fsparser.cpp
 	src/mcsParser.cpp
 	src/ftdispi.cpp
@@ -139,7 +138,6 @@ set(OPENFPGALOADER_HEADERS
 	src/ice40.hpp
 	src/ihexParser.hpp
 	src/progressBar.hpp
-	src/pathHelper.hpp
 	src/rawParser.hpp
 	src/usbBlaster.hpp
 	src/bitparser.hpp
@@ -198,6 +196,9 @@ target_link_libraries(openFPGALoader
 if (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
 	# winsock provides ntohs
 	target_link_libraries(openFPGALoader ws2_32)
+	
+	target_sources(openFPGALoader PRIVATE src/pathHelper.cpp)	
+	list(APPEND OPENFPGALOADER_HEADERS src/pathHelper.hpp)
 endif()
 
 # libusb_attach_kernel_driver is only available on Linux.
--- a/src/altera.cpp
+++ b/src/altera.cpp
@@ -13,8 +13,10 @@
 #include "device.hpp"
 #include "epcq.hpp"
 #include "progressBar.hpp"
-#include "pathHelper.hpp"
 #include "rawParser.hpp"
+#if defined (_WIN64) || defined (_WIN32)
+#include "pathHelper.hpp"
+#endif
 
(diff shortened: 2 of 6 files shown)


trabucayre:
LGTM.
If this fix #262 I will apply this.

gregdavill:
In theory this fixes the path issue. But it's also breaking the executable.

Let me try applying the path change before it's compiled in.

trabucayre:
This approach looks good but I hesitate between this and using inconditionaly a relative path when openFPGALoader is build using mingw.

gregdavill:
I agree a relative path would be nicer.

But simply adjusting the path to `../share/openFPGALoader/spiOverJtag_` doesn't work. As the call to fopen() parses this based on the current working directory, not the binary location. So it only finds the files if you're already in `mingw64/bin/`

trabucayre:
So relative path is a bad option (or it's required to update altera.cpp and xilinx.cpp to check path type and to do a live update: I'm not sure it's a relevant solution).
So long term solutions are the both introduces in #262

gregdavill:
My initial understanding of how paths work in mingw was wrong. 
We do require a full absolute path starting with `C:\...` but we also don't want to bake in the CI's environment into that path.

So on windows builds there is a non-absolute path bake into `DATA_DIR` `/mingw64/share/`.
At runtime we call `cygpath -m` to create an absolute path from this. 

```
$ openFPGALoader -b qmtechKintex7 -c tigard -f  fpga.bit
write to flash
Board qmtechKintex7 has not default cable
Open file DONE
Parse file DONE
use: C:/msys64/mingw64/share/openFPGALoader/spiOverJtag_xc7k325tffg676.bit.gz
load program
Done
Detected: spansion S25FL256L 512 sectors size: 256Mb
00000000 00000000 00000000 00
Done
Done
```

CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ set(OPENFPGALOADER_SOURCE
 	src/jlink.cpp
 	src/lattice.cpp
 	src/progressBar.cpp
+	src/pathHelper.cpp
```
trabucayre, line 114: This class is currently only used when openFPGALoader is built for windows: maybe it must be conditionally added ?

CMakeLists.txt
```diff
@@ -138,6 +139,7 @@ set(OPENFPGALOADER_HEADERS
 	src/ice40.hpp
 	src/ihexParser.hpp
 	src/progressBar.hpp
+	src/pathHelper.hpp
```
trabucayre, line 142: ditto

src/altera.cpp
```diff
@@ -13,6 +13,7 @@
 #include "device.hpp"
 #include "epcq.hpp"
 #include "progressBar.hpp"
+#include "pathHelper.hpp"
```
trabucayre, line 16: ditto

trabucayre:
LGTM!

CMakeLists.txt
gregdavill: That's a good idea

trabucayre:
Applied!
Thanks!

PR #264, opened 2022-10-25
improve svf parsing.

phdussud:
I want to be able to parse and execute Lattice SVF files produced by Diamond. This still isn't a complete support for all SVF specifications but it is a step forward.

### DIFF
--- a/src/svf_jtag.cpp
+++ b/src/svf_jtag.cpp
@@ -2,12 +2,13 @@
 /*
  * Copyright (C) 2019 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
  */
-
+#include <algorithm>
 #include <iostream>
 #include <sstream>
 #include <fstream>
 #include <map>
 #include <vector>
+#include <unistd.h>
 
 #include "jtag.hpp"
 
@@ -18,8 +19,8 @@ using namespace std;
 void SVF_jtag::split_str(string const &str, vector<string> &vparse)
 {
 	string token;
-	std::istringstream tokenStream(str);
-	while (std::getline(tokenStream, token, ' '))
+	istringstream tokenStream(str);
+	while (getline(tokenStream, token, ' '))
 		vparse.push_back(token);
 }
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Hi,
Thanks for this PR. (and thanks to fix some details from this early code).
Since it make many modifications I have to take time to check everything.
A quick note: some `if` statements have `{` in a new line: to be coherent with the rest of the code could you fix that (ie the `{` just after condition to the same line)?

phdussud:
I think I caught all of the offending { placements. I changed the behavior of my editor to respect this style so it shouldn't happen in the future.

trabucayre:
Sorry for delay to review...

Applied!
Thanks!

phdussud:
Thanks! Now that this is done, I would like to ask you a question. why is SVF support limited to a small number of chips? SVF is meant to be universal, it is just a way to specify JTAG signals. I wanted to program an ECP5 FPGA with SVF. 
Would you accept a change that detect a SVF file through its name or type and play it to the selected JTAG target? It would render the SVF specific code redundant on the chip types who currently support SVF, since it does nothing more than play the SVF

trabucayre:
Of course: adding svf support for all device type make sense but I have never doing that because it's a bit slower than using native protocol: official SVF specification has no notion of loop so you have to wait longest possible duration instead of polling for  a status bit.
In fact at the beginning of this project after having implemented the jtag FSM, I have added svf format to play with my trenz cyc1000 (intel/altera cyclone 10). Second board/FPGA supported was the arty (arty), but for this support I have read Xilinx's programming manual where the sequence is explained and obtained a native support. Third FPGA was the machXO3 and again Lattice provides programming sequence ->again no need to adapts/uses SVF file.

Now for lattice/gowin/xilinx/efinix/colognechip/... quite all family are supported and SVF filetype is more or less useless. Intel remain the only one where some family (max10) needs to uses SVF.

phdussud:
I see what you are saying. One argument for adding SVF is that you can program parts that are not supported by OpenFGALoader. 
The other part is that it isn't easy to convert a SVF file to a bit file. Since the bulk of the work has been done, and we could remove the redundant code for specific models, it would make the code base smaller.

trabucayre:
> I see what you are saying. One argument for adding SVF is that you can program parts that are not supported by OpenFGALoader.

It's exactly situation for intel *max10*: native support is not (currently) available. But native support is always better because faster. For example when you have something like
```
RUNTEST 128 TCK;
SDR 32 TDI (00000000) TDO (1E000000) MASK (FF800000);
```
It mean you have to poll a register until content is equal to a specific value, the number of TCK is related to the longest duration and is based on the the clock frequency specified at svf generation.

> The other part is that it isn't easy to convert a SVF file to a bit file. 

If you have only the SVF it's true, but if you have the gateware: the SVF is usually an extra operation (or an option to set) to convert native bitstream file to SVF file

> Since the bulk of the work has been done, and we could remove the redundant code for specific models, it would make the code base smaller.

I'm not sure to understand this sentence: it's to have a specific behaviour based on the file extension and independant to the target? Or it's to reduce native support and to move to an SVF player only?

phdussud:
Specifically, I would like to remove every duplicated piece of code that detects svf files in files dedicated to specific parts and add this code (or wrap into a function) at line main.cpp line 492 (after jtag->device_select(index)) :
`
	

	if (!args.bit_file.empty() && 
		((!args.file_type.empty() && args.file_type == "svf") ||
		 (args.bit_file.substr(args.bit_file.find_last_of(".") + 1) == "svf"))) {
		SVF_jtag *svf = new SVF_jtag(jtag, args.verbose);
		try {
			svf->parse(args.bit_file);
		} catch (std::exception &e) {
			return EXIT_FAILURE;
		}
		return EXIT_SUCCESS;
	}`

trabucayre:
This look fine. I'm agree. SVF has don't care about device and need only to have access to the JTAG level.
I see two possibilities: a new option `--svf` as done for **dfu** or **xvc** or simply having a `configBitstreamParser` classmethod to extract file extension and to use it near the begin of jtag part (I prefer a bit the second option to avoid a new CLI option).

phdussud:
I like the second option as well. Just to make sure, the `configBitStreamParser` class method body will do exactly this: 
`(!args.bit_file.empty() && 
	((!args.file_type.empty() && args.file_type == "svf") ||
	 (args.bit_file.substr(args.bit_file.find_last_of(".") + 1) == "svf")))`
Is this what you have in mind?

trabucayre:
In fact it's a mistake it's `device.cpp` but since it's just a find and a substr so it's not really relevant to use a class method...

PR #265, opened 2022-10-29
Update boards.yml - added Alchitry Au+

devdc:
Tested on Alchitry Au+ and works well on macOS Monterey

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -19,7 +19,14 @@
   FPGA: Artix xc7a35tftg256
   Memory: OK
   Flash: OK
-
+  
+- ID: alchitry_au_plus
+  Description: Alchitry Au+ (Plus)
+  URL: https://www.sparkfun.com/products/17514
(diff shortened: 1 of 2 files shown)


trabucayre:
Thanks for this PR but could you add entry near [here](https://github.com/trabucayre/openFPGALoader/blob/master/src/board.hpp#L107)? Something like
```
JTAG_BOARD("alchitry_au_plus",     "xc7a100tftg256",  "ft2232",   0, 0, CABLE_DEFAULT),
```
Since it's not the same model (size) than the `alchitry_au`, with this new entry you could use `-b alchitry_au_plus` to have everything working out of the box.
Thanks!

devdc:
Thanks! hopefully I was able to follow your instructions :)

trabucayre:
Applied!
Thanks!

devdc:
Thank YOU and thanks for creating and maintaining that amazing project

PR #269, opened 2022-11-10
Removed the svf programming code from devices and added device independent svf programming

phdussud:
… in main.cpp

Fixed a type error in svf_jtag.cpp

### DIFF
--- a/src/svf_jtag.cpp
+++ b/src/svf_jtag.cpp
@@ -42,8 +42,8 @@ static unsigned char *parse_hex(string const &in, size_t byte_length,
 {
 	unsigned char *txbuf = new unsigned char[byte_length];
 	char c;
-	size_t last_iter = in.size() - (2 * byte_length);
-	for (size_t i = (in.size() - 1), pos = 0; i >= last_iter; i--, pos++) {
+	ssize_t last_iter = in.size() - (2 * byte_length);
+	for (ssize_t i = (in.size() - 1), pos = 0; i >= last_iter; i--, pos++) {
 		if (i < 0) {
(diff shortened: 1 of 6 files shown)


src/svf_jtag.cpp
```diff
@@ -42,8 +42,8 @@ static unsigned char *parse_hex(string const &in, size_t byte_length,
 {
 	unsigned char *txbuf = new unsigned char[byte_length];
 	char c;
-	size_t last_iter = in.size() - (2 * byte_length);
-	for (size_t i = (in.size() - 1), pos = 0; i >= last_iter; i--, pos++) {
+	ptrdiff_t last_iter = in.size() - (2 * byte_length);
+	for (ptrdiff_t i = (in.size() - 1), pos = 0; i >= last_iter; i--, pos++) {
```
trabucayre, line 46: Why using a `ptrdiff_t`: this type is used with pointers. `std::string::size()` return a `size_t` and `byte_length` is also a `size_t`

trabucayre:
LGTM thanls!
Just the ptrdiff_t detail

phdussud:
That's because last_iter can be negative. I thought that the signed equivalent (in terms of size) of size_t is ptr_diff_t

trabucayre:
`ssize_t` seems a better candidate

> [...]
> According to POSIX, it shall be a signed integer type capable of storing values at least in the range [-1, SSIZE_MAX]
> [...]

phdussud:
Yes, that's the one I should be using!

trabucayre:
LGTM.
Applied.
Thanks!

PR #275, opened 2022-11-26
ALINX AX516 board.

inkdot7:
I'd like to report success in programming an ALINX AX516 Spartan-6 FPGA Development Board.

Programming just worked out of the box!  Thanks!

./openFPGALoader -v -v -c ft232 someproj.bit

gives the following output:

```
found 1 devices
index 0:
	idcode 0x4002093
	manufacturer xilinx
	family spartan6
	model  xc6slx16
	irlength 6
File type : bit
```

I am programming it with an ALINX AL321 USB cable.  (Since external, I did not specify ft232 it in `board.hpp`, is that correct?)

Writing to flash also works (now with `-b alinx_ax516`), and tells:

```
Detected: Winbond W25Q128 256 sectors size: 128Mb
```

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -27,6 +27,13 @@
   Memory: OK
   Flash: OK
   
+- ID: alinx_ax516
+  Description: ALINX AX516
+  URL: https://www.alinx.com/en/detail/281
+  FPGA: Spartan6 xc6slx16csg324
+  Memory: OK
+  Flash: OK
+
 - ID: arty_a7_35t
   Description: Digilent Arty A7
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM.
Applied Thanks!

trabucayre:
If you haven't to explicitly specify cable I assume it's based on ft2232.
I have one alinx cable: I have to check if it's the same.

inkdot7:
Just for clarity, I have to specify `-c ft232` also when writing to flash.

trabucayre:
Ok. Thank for this precision. The only situation where the cable name may be omitted, for a board without a default cable, is when it's based on a ft2232.

PR #280, opened 2022-12-04
Add support for Xc6slx9tqg144 FPGA

saursin:
Hello,
I've tried to add support for Xilinx Xc6slx9tqg144 FPGA Part. 
earlier, calling `openFPGALoader --list-fpga` used to show that `xc6slx9` part is supported but the `spiOverJtag.bit.gz` file was missing for the same. I've added this file along with some tweaks referring to c38d55e.

Thanks


trabucayre:
LGTM. Just a small remark: `.bit` files was integrated before adding *gzip* support and left to avoid to many noise by removing/replacing `.bit` by `.bit.gz`. Now only compressed version must be added.
Could you remove this .bit file `amend` your commit and force push?
Thanks.

trabucayre:
Applied!
Thanks!

saursin:
Hmm.. looks like merge happened as soon as I reverted my changes, (because I didn't do a force-push earlier). and therefore the .bit file got pushed to master.

trabucayre:
Don't worry: I have removed the file. Both action may have been done at same time :)

PR #283, opened 2022-12-11
Add initial support for the Xilinx VCU118 board

barbedo:
Hey,

Here's a basic support for the Xilinx VCU118 board / Virtex UltraScale+ XCVU9P FPGA.
I could test it locally with a complex design and it works as expected.

The `spiOverJtag` image is missing.
I'll try to add it later in a new PR when I have access to a machine with Vivado and the proper (paid) license.

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -203,6 +203,13 @@ Xilinx:
     Memory: OK
     Flash: NA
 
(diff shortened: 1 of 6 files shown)


trabucayre:
Applied!
Thanks!
Yes it's always complex when a license is required to produces bitstreams :-/

PR #286, opened 2022-12-25
feat(board): add support for Trenz C10LP-RefKit

boogermann:


trabucayre:
Applied Thnaks!
Have you tried flash? (it's requires an update to spiOverJtag)?

PR #289, opened 2023-01-04
feat(board): add support for Arrow/Terasic DECA

boogermann:


trabucayre:
Applied thanks @boogermann !

PR #292, opened 2023-01-19
bitparser: check bitstream length in the file

michalsieron:
It was present before, but for some reason was removed in 16932786.

Signed-off-by: Michal Sieron <msieron@antmicro.com>

### DIFF
--- a/src/bitparser.cpp
+++ b/src/bitparser.cpp
@@ -110,7 +110,7 @@ int BitParser::parse()
 	int pos = parseHeader();
 
 	/* _bit_length is length of data to send */
-	int rest_of_file_length = _raw_data.size() - pos;
+	int rest_of_file_length = _file_size - pos;
 	if (_bit_length < rest_of_file_length) {
 		printWarn("File is longer than bitstream length declared in the header: " +
 				std::to_string(rest_of_file_length) + " vs " + std::to_string(_bit_length)
--- a/src/bitparser.cpp
+++ b/src/bitparser.cpp
@@ -109,10 +109,21 @@ int BitParser::parse()
 	/* process all field */
 	int pos = parseHeader();
 
-	/* rest of the file is data to send */
-	_bit_data.resize(_raw_data.size() - pos);
-	std::move(_raw_data.begin() + pos, _raw_data.end(), _bit_data.begin());
-	_bit_length = _bit_data.size();
+	/* _bit_length is length of data to send */
+	int rest_of_file_length = _raw_data.size() - pos;
+	if (_bit_length < rest_of_file_length) {
+		printWarn("File is longer than bitstream length declared in the header: " +
+				std::to_string(rest_of_file_length) + " vs " + std::to_string(_bit_length)
+		);
+	} else if (_bit_length > rest_of_file_length) {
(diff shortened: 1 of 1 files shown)


trabucayre:
Good catch!
LGTM just few remark (see bellow)

src/bitparser.cpp
```diff
@@ -109,10 +109,21 @@ int BitParser::parse()
 	/* process all field */
 	int pos = parseHeader();
 
-	/* rest of the file is data to send */
-	_bit_data.resize(_raw_data.size() - pos);
-	std::move(_raw_data.begin() + pos, _raw_data.end(), _bit_data.begin());
-	_bit_length = _bit_data.size();
+	/* _bit_length is length of data to send */
+	int rest_of_file_length = _raw_data.size() - pos;
```
trabucayre, line 113: you can use `_file_size` (see configBitstreamParser)

src/bitparser.cpp
```diff
@@ -109,10 +109,21 @@ int BitParser::parse()
...
+		);
+	} else if (_bit_length > rest_of_file_length) {
+		printError("File is shorter than bitstream length declared in the header: " +
+				std::to_string(rest_of_file_length) + " vs " + std::to_string(_bit_length)
+		);
+		return 1;
+	}
+
+	_bit_data.resize(_bit_length);
+	std::move(_raw_data.begin() + pos, _raw_data.begin() + pos + _bit_length, _bit_data.begin());
```
trabucayre, line 126: here `_bit_length == _file_size - pos` so `pos + _bit_length == pos + _file_size - pos == _file_size` ie it's the same thing to keep `end()`.

src/bitparser.cpp
trabucayre: sorry `_file_size - pos`

src/bitparser.cpp
michalsieron: But `_bit_length == _file_size - pos` doesn't have to be always true.

Assume you have a bitstream file with extra data appended after the actual bitstream data.
So you have `header|bitstream|extra data`.

If we do `_raw_data.end()` we copy this `extra data` as well, even though field `(e)` specifies its length and it doesn't include the `extra data` appended later.

src/bitparser.cpp
michalsieron: So I just need to replace `_raw_data.size()` with `_file_size`?

src/bitparser.cpp
trabucayre: It's true! My fault! I haven't seen there is no return when `_bit_length < rest_of_file_length`.
In this situation using `begin() + pos + _bit_length` instead of `end()` is the solution.

src/bitparser.cpp
trabucayre: Exactly. This is done in the [configBitstreamParser class](https://github.com/trabucayre/openFPGALoader/blob/master/src/configBitstreamParser.cpp) lines 54 or 73 or 87

michalsieron: Done

trabucayre:
Applied! Thanks @michalsieron

Issue #293, opened 2023-01-21
Add flash support for the Xilinx VCU118 board


barbedo:
Hey again,

This is a followup from #283, since I've got access to a full Vivado license.

Adding support for the dual QSPI flash on this board was more difficult than I had expected, so I'll try to document all the modifications and the rationale behind them here and break down the code modifications in multiple PRs to ease a bit the review.

If these modifications are merged, it can potentially open the door to support other Xilinx UltraScale-based boards with Dual QSPI Flash, like the KCU105, KCU1116, VCU110, ZCU102, etc.

### Changes

- Add support for the MT25QU01 QSPI flash chip.

- Add support for flash chips with more than 128 Mb by adding 4-byte addressing when writing to flash addresses higher than `0xFFFFFF`.
    - I decided to do it dynamically by using the 4-byte address variation of the read/write/erase commands only if the address is higher than 128 Mb. From what I could find online, pretty much every chip from Micron/Infineon/Spansion with more than 128 Mb supports it.
- Modify the `spiOverJtag` module to take into account the new differences introduced by the UltraScale series.
    - The pads to the primary QSPI flash are not accessible anymore directly from the design and need to be accessed through the `STARTUPE3` module.
    - The pads to the secondary QSPI flash are accessible as normal I/Os after configuration, except for CCLK, which is shared with the primary flash chip.
    - To avoid maintaining two bitfiles, each one targeting a different flash chip, I chose instead to select which QSPI flash to use based on the boundary scan user instruction that is selected: `USER1` for the first flash, `USER2` for the second flash.
    - Two `BSCANE2` blocks are now instanced.
    - I tried to modify the design only if the `virtexultrascale` flag is present.
    - See below the diagram for the design:
![vcu118-dual-qspi-flash](https://user-images.githubusercontent.com/13098357/213870279-0fcf08e8-3041-4bb3-bed1-b140fafca950.png)
    - I regenerated the design for the Arty A7 100T (the only other board that I have access to right now) to verify that it didn't break the bitfile for the other Xilinx FPGAs.

- Make the instruction register codes variable according to the Xilinx FPGA being used.
    - Only `default` or `xcvu9p` for now.
    - According to _UltraScale Architecture Configuration User Guide UG570_ Table 1-5, the instruction register lengths are variable, because of their 3D IC thingy. The FPGA in the VCU118 board has an IR length of 18 bytes.
[...]

PR #294, opened 2023-01-21
Add support for MT25QU01 flash

barbedo:
Refer to #293.


trabucayre:
Applied Thanls @barbedo !

PR #295, opened 2023-01-21
Add support for 4-byte QSPI flash addressing

barbedo:
Refer to #293.

### DIFF
--- a/src/spiFlash.cpp
+++ b/src/spiFlash.cpp
@@ -3,6 +3,7 @@
  * Copyright (C) 2019 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
  */
 
+#include <stdint.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


src/spiFlash.cpp
```diff
@@ -107,35 +119,77 @@ int SPIFlash::bulk_erase()
...
+		tx[0] = static_cast<uint8_t>(FLASH_4SE          );
+		tx[1] = static_cast<uint8_t>(0xff & (addr >> 24));
+		tx[2] = static_cast<uint8_t>(0xff & (addr >> 16));
+		tx[3] = static_cast<uint8_t>(0xff & (addr >>  8));
+		tx[4] = static_cast<uint8_t>(0xff & (addr      ));
+		len = 5;
+	}
+
+	_spi->spi_put(tx, NULL, len);
+
```
trabucayre, line 141: It's looks good but too avoid to many dupplication maybe it's better to do something like (and of course same idea for next modifications):

```cpp
    uint8_t tx[5];
    uint32_t len = 0;

    uint8_t cmd = (addr <= 0xffffff) ? FLASH_SE : FLASH_4SE;

    if (addr <= 0xffffff) {
        tx[len++] = static_cast<uint8_t>(cmd                );
        if (cmd == FLASH_4SE)
            tx[len++] = static_cast<uint8_t>(0xff & (addr >> 24));
        tx[len++] = static_cast<uint8_t>(0xff & (addr >> 16));
        tx[len++] = static_cast<uint8_t>(0xff & (addr >>  8));
        tx[len++] = static_cast<uint8_t>(0xff & (addr      ));

    _spi->spi_put(tx, NULL, len);

```

barbedo: I completely agree. Please check the updated version.

To verify that this still works, I've cherry-picked this latest commit on top of my `vcu118-flash` branch and loaded/dumped a memory file that is larger than 16 MB.

PR #296, opened 2023-01-21
Add XCVU9P spiOverJtag bitfile with SPIx8 support

barbedo:
Refer to #293.


trabucayre:
LGTM.
Just a remark: why updating xc7a100t's bitstream? Since modifications are conditioned by `define` the behaviour remains the same for all others devices?

barbedo:
Thanks!

I've regenerated and tested the Arty A7 100T bitfile to verify that I didn't break the logic outside the `define`.

I've included it in the commit just to "prove" that it still works. The behavior is exactly the same as the previous one.

If you don't want this binary in the master's history, I'll remove it.

trabucayre:
In fact I try, as much as possible, to limit remove/update about spiOverJtag, to avoid increasing repository size (it's why some bitstream are uncompressed).
Discarding this file to this commit requires:
```
git rm thefile
git commit --amend
git push -f
```
As you want. If you prefer not to try these commands: I will apply with this bitstream.

barbedo:
I completely agree with the concern over the repo size impact.

I've just removed the Arty's bitstream.

trabucayre:
Applied thanks!

PR #297, opened 2023-01-21
Add support for XCVU9P IR length and codes

barbedo:
Refer to #293.

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -680,6 +680,6 @@
 - ID: vcu118
   Description: Xilinx VCU118
   URL: https://www.xilinx.com/products/boards-and-kits/vcu118.html
-  FPGA: Virtex UltraScale+ xcvu9pl2flga2104e
+  FPGA: Virtex UltraScale+ xcvu9p-flga2104
   Memory: OK
   Flash: NA
--- a/src/board.hpp
+++ b/src/board.hpp
@@ -209,7 +209,7 @@ static std::map <std::string, target_board_t> board_list = {
 	JTAG_BOARD("papilio_one",     "xc3s500evq100", "papilio", 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("usrpx300",        "xc7k325tffg900", "digilent", 0, 0, CABLE_MHZ(15)),
 	JTAG_BOARD("usrpx310",        "xc7k410tffg900", "digilent", 0, 0, CABLE_MHZ(15)),
-	JTAG_BOARD("vcu118",          "xcvu9pl2flga2104e", "jtag-smt2-nc", 0, 0, CABLE_DEFAULT)
+	JTAG_BOARD("vcu118",          "xcvu9p-flga2104", "jtag-smt2-nc", 0, 0, CABLE_DEFAULT)
 };
 
 #endif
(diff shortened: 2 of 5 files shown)


trabucayre:
The principle looks good, but my worry is about a potential global slow down when using xilinx devices?
The same solution as for zynqmp isn't working (ie injecting a dummy device into the chain)?

barbedo:
I wanted to go through the same code path as the 7-series implementation, and avoid doing ad-hoc or hacky stuff.

The manual says that the device has a single TAP (which is a little different from the ZynqMP situation with PL TAP+PS TAP in the chain), but the IR length is variable depending on the part being used.

So, for the XCVU9P, I need to treat it as a single device with an 18 bit long IR and shift all the values when writing to it.

If by slowdown, you mean the fact that:

```c
_jtag->shiftIR(JSHUTDOWN, 6);
```

is being replaced by:

```c
_jtag->shiftIR(get_ircode(_ircode_map, "JSHUTDOWN"), NULL, _irlen);
```

The first version of `shiftIR` is calling the second one anyway, so the only overhead is the access to the `ircode_mapping` map, which should be fast, since it is a hash table.

Let me benchmark this version with the latest release on the Arty, just to be sure.

barbedo:
Here are some results while loading a relatively complex design (~3.8MB of flash data) to an Arty A7 100T, compiling on a macOS M1 machine:

### On `master`
```
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.94s user 6.83s system 18% cpu 46.972 total
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.94s user 7.00s system 19% cpu 47.051 total
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.90s user 6.30s system 17% cpu 46.573 total
```

### On `barbedo:/vcu-flash-ir-codes`
```
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.97s user 6.56s system 18% cpu 47.142 total
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.95s user 6.33s system 17% cpu 46.729 total
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.95s user 6.52s system 18% cpu 46.693 total
```

So there seems to be a small ~2% time increase in userspace. This could be jitter, but I think that this really comes from the additional memory accesses when writing to the IR. Hopefully, this is not too much impact for you.

One question that just came to mind. I don't see the optimization flags being set anywhere in the repo. Are the releases compiled in `-O0`?

By manually adding `-O2`, I have a little speed increase (~16%) for this same test:
```
./build/openFPGALoader -f --board arty_a7_100t --bitstream  --bridge   --freq  1.64s user 6.57s system 17% cpu 46.598 total
```

trabucayre:
Ok. Timings differences seems in the error bar. Maybe it's more visible with an highest frequency (ie 30e6) but it's acceptable I assume.
I haven't added -O3 and let compiler decide but maybe it make sense to force this optimization level or to explain how to specify that at cmake step (or adding -03 and explain how to reducing this level).

trabucayre:
Applied thanks!

PR #298, opened 2023-01-21
Add target-flash and secondary-bitstream CLI options for VCU118

barbedo:
Refer to #293.

~Depends on #297, so mark as draft for now.~ #297 is now merged, so opening for review.

### DIFF
--- a/src/xilinx.cpp
+++ b/src/xilinx.cpp
@@ -20,6 +20,7 @@
 #include "rawParser.hpp"
 
 #include "display.hpp"
+#include "spiInterface.hpp"
 #include "xilinx.hpp"
 #include "xilinxMapParser.hpp"
 #include "part.hpp"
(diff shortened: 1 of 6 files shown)


trabucayre:
Not finished to review this PR but in one hand you use multiple time `target_flash` as an *string* and in another hand you have defined an *enum*: why not fixing `PRIMARY_FLASH = 1`, `SECONDARY_FLASH = 2` and `target_flash` as this enum type?
So this allows to do something like
```
if (xxx="both" || xxx="primary)
  target_flash = PRIMARY
if (xxx="both" || xxx="secondary)
  target_flash |= SECONDARY
```

and every time you have to use this var:
```
if (target_flash & PRIMARY)
// do something with primary flash
if (target_flash & SECONDARY)
// do something with secondary flash
```

barbedo:
Hm, I wanted to separate the concepts of:
- The command line arguments (`primary`, `secondary`, `both`), owned by `main.cpp` and propagated to the other classes (only Xilinx for now/ever I guess).
- From the "which flash chip is being targeted by the specific implementation of dual-QSPI for Xilinx UltraScale devices", owned by `xilinx.cpp`.

But if it makes more sense to you, I can create a `xilinx_flash_chip_t flash_chips` in `main()` and populate its bitfields based on the strings before passing it to the Xilinx class, like you propose.

Just give me the go and I'll update the PR.

trabucayre:
My proposition/idea is not clear in fact. `main.cpp` modifications looks good to me: this file has nothing (as must as possible) to know about vendors specifics things. In fact when I talk about using the `enum` with a bitmask instead of comparision with string is internally to `xillinx.cpp`.

barbedo:
Hey, sorry for the delay.

I've kept the CLI options as strings on `main` and converted them to a bitfield that uses the `xilinx_flash_chip_t` enum in the Xilinx constructor. Then I can use the bitfield as you suggested on each flash access.

The builds are failing though:
```
E: Failed to fetch [...].deb  Could not connect to azure.archive.ubuntu.com:80 (40.119.46.219), connection timed out [IP: 40.119.46.219 80]
```

src/xilinx.cpp
```diff
@@ -334,32 +377,27 @@ void Xilinx::program(unsigned int offset, bool unprotect_flash)
...
-		return;
-	}
-
-	printSuccess("DONE");
-
-	printInfo("Parse file ", false);
-	if (bit->parse() == EXIT_FAILURE) {
 		printError("FAIL");
 		delete bit;
+		delete secondary_bit;
```
trabucayre, line 391: when `bit` fails `secondary_bit` is a `nullptr` maybe (and in fact for both) a test is required before using `delete` ?

trabucayre:
LGTM. Just this small test to add. Thanks!

src/xilinx.cpp
barbedo: I've just added a check.

trabucayre:
Applied Thanks!

trabucayre:
Thanks for your contributions too !
Yep, I usually miss to update/improve documentation too :-/

PR #301, opened 2023-01-29
part: Add all cyclone V variant, board: Add the Terasic C5G

galibert:


trabucayre:
LGTM. But you needs to update [doc](https://github.com/trabucayre/openFPGALoader/blob/master/doc/boards.yml), also device ref may be useful to write bitstream info flash.

If you wish to write bitstream into flash you have to add an spiOverJtag variant

Thanks!

galibert:
Ok, I'll try flash.  Not sure if I dare changing it... I need to be sure I have the original one handy to reflash it, because it is quite useful for testing.

trabucayre:
Don't worry for flash side. It's may be another PR in a future.
Just update boards.yml and add FPGA code and this will be good to me.
Thanks

trabucayre:
Applied thanks!

PR #308, opened 2023-02-14
Add documentation for Xilinx SPIx8 mode

barbedo:
Should solve #293.

Let me know if this level of information is ok.


trabucayre:
Applied Thanks!

PR #310, opened 2023-02-15
Fix typo in part name and family.

bl0x:
I found wrong information on the web when looking up the part name from the ID code.
Sorry about that.


trabucayre:
Thanks applied!
I searched too and seen some ref for xa2 matching coolrunner2: maybe really old FPGA code :)

PR #313, opened 2023-02-19
Add initial support for VCU128

jiegec:
Add bitstream loading to SRAM support for VCU128 board. Most parts are the same as VCU118 except for ftdi chip and IDCODE.

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -704,3 +704,10 @@
   FPGA: Virtex UltraScale+ xcvu9p-flga2104
   Memory: OK
   Flash: OK
+
+- ID: vcu128
+  Description: Xilinx VCU128
+  URL: https://www.xilinx.com/products/boards-and-kits/vcu128.html
+  FPGA: Virtex UltraScale+ xcvu37p-fsvh2892
+  Memory: OK
+  Flash: NA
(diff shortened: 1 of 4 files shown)


trabucayre:
Applied! Thanks!
Have you tried to program the flash ?

jiegec:
> Have you tried to program the flash ?

Working on it! I am trying to understand how programing the flash works(seems implemeneted in `spiOverJtag` directory), but I could not find docs.

trabucayre:
Yes. I know documention about how to contribute is missing and there is more or less nothing about `spiOverJtag`.
[this commit](https://github.com/trabucayre/openFPGALoader/commit/29d2264e1594b279b49dc883b1f35fa8652efa6e) is good stating point since it address same FPGA family.

In fact in your specific case:

- no need to modify *verilog* file since the family is supported
- if *package* is not already present you have to create `xdc` file
- `Makefile` must be updated to add your target device 
- `run.py` must also be updated to extract *family*, *size* and *package*

jiegec:
> Yes. I know documention about how to contribute is missing and there is more or less nothing about `spiOverJtag`. [this commit](https://github.com/trabucayre/openFPGALoader/commit/29d2264e1594b279b49dc883b1f35fa8652efa6e) is good stating point since it address same FPGA family.
> 
> In fact in your specific case:
> 
> * no need to modify _verilog_ file since the family is supported
> * if _package_ is not already present you have to create `xdc` file
> * `Makefile` must be updated to add your target device
> * `run.py` must also be updated to extract _family_, _size_ and _package_

Thanks! I successfully built a bitstream file and can program the flash! However, if failed verification i.e. readback != written.

PR #316, opened 2023-02-20
Add flash support for VCU128

jiegec:
VCU128 does not have secondary flash, so an additional ifdef check is done for VCU118.


trabucayre:
LGTM, but maybe a check at `xilinx.cpp` level may make sense to throw an exception when user try to write to the secondary flash for a device with only one. It may also simplify `xilinx_spiOverJtag.v` by simply conditionting everything about secondary flash to peripherals with two chips?
Thanks

jiegec:
> LGTM, but maybe a check at `xilinx.cpp` level may make sense to throw an exception when user try to write to the secondary flash for a device with only one. It may also simplify `xilinx_spiOverJtag.v` by simply conditionting everything about secondary flash to peripherals with two chips? Thanks

Done.

trabucayre:
Applied thanks!

PR #321, opened 2023-02-28
Opal Kelly xem8320 board info

AEW2015:


trabucayre:
Thanks for this PR but seems I was too fast (or my mail client too slow).
I have added this board [here](https://github.com/trabucayre/openFPGALoader/commit/7de494f66635cb36b4227467f8f015d69341b142).
Thanks!

PR #322, opened 2023-03-01
Add Support for Xilinx KCU116

mander1000:
Hey

I added support for the Xilinx KCU116 board (Kintex Ultrascale Plus).

I tested it to flash both SPI flash chips on the board (QSPI), which worked fine.

Feel free to comment and let me know if you need anything else.

mander


spiOverJtag/build.py
```diff
@@ -127,15 +131,22 @@
...
                     'default': 1}
             elif part == "xcvu37p-fsvh2892":
                 tool_options = {'part': part + '-2L-e'}
+            elif part == "xcku5p-ffvb676":
+                tool_options = {'part': part + '-2-e'}
+                parameters["secondaryflash"]= {
+                    'datatype': 'int',
+                    'paramtype': 'vlogdefine',
+                    'description': 'secondary flash',
+                    'default': 1}
```
trabucayre, line 149: I'm not familiar about speedgrade for ultrascale+ but `-1` or `-2` are similar to equivalent for artix (for example)?
If so is it possible to use a low speed grade with a FPGA with highest grade?
The idea is to merge xcvu and xcku in the same test.
If I have true could you modify it (no need to rebuild the bitstream).
Thanks

mander1000: The naming for the speed grade is the same on all UltraScale devices (and I think even on all Xilinx devices). `-1` is the slowest. So sure, I changed that. I combined the family name to `Xilinx UltraScale` as it checks for `Virtex` and `Kintex` devices. Hope that is ok.

mander1000:
should all be done now and ready to merge. :)

trabucayre:
Applied. Thanks!

PR #327, opened 2023-04-07
Add support for xc7a25t FPGA

CreatingNull:
Really great tool, thanks for all your work developing this.

I've added support for the Artix-7 a25t. 
I've tested this the JTAG->SPI flashing works on the `csg325` package, but I don't have the other `cpg238` device to confirm that binary.

Let me know if I've missed anything.


trabucayre:
LGTM! Thanks!
A small thing: only compressed files must be versioned. Could you remove .bit files, preferably using:
- `git rm spiOverJtag_xc7a25tcsg325.bit spiOverJtag_xc7a25tcpg238.bit`
- `git commit --amend`
- `git push -f`
This has for goal to remove reference to both .bit without increasing repository size. I know there is some .bit files in `spiOverJtag` directory but these one was versioned before adding gzip support.
Thanks again

CreatingNull:
> LGTM! Thanks! A small thing: only compressed files must be versioned. Could you remove .bit files, preferably using:
> 
>     * `git rm spiOverJtag_xc7a25tcsg325.bit spiOverJtag_xc7a25tcpg238.bit`
> 
>     * `git commit --amend`
> 
>     * `git push -f`
>       This has for goal to remove reference to both .bit without increasing repository size. I know there is some .bit files in `spiOverJtag` directory but these one was versioned before adding gzip support.
>       Thanks again

That makes sense, I was trying to follow what was present for other devices from the same family.
I've removed the .bit files.

trabucayre:
Applied! Thanks @CreatingNull

PR #328, opened 2023-04-09
Fix `cmsisdap` support on macOS

chenzhuoyu:
Building openFPGALoader on macOS does not enable `cmsisdap`, even if `hidapi` was installed through Homebrew.

After some tinkering, I found some issues within the `CMakeLists.txt`:

1. `link_directories` should be placed **before** the call to relevant `add_executable` in order to be effective, otherwise it won't work, and the library path won't be added to the linker's search path.

    The document doesn't mention that (or I can't find one), but there is an answer on SO said just that (https://stackoverflow.com/a/40554704).

    This PR fixes the issue mentioned above.

2. The original `CMakeLists.txt` does not add library paths for `hidapi`, this PR also fixes that.

3. macOS Homebrew installs `hidapi` as just `hidapi` instead of `hidapi-hidraw` or `hidapi-libusb`, so I also added the relavent `pkg_check_modules` for it.

Fixing these issues will make `cmsisdap` available for macOS.


trabucayre:
Applied. Thanks!

PR #331, opened 2023-04-16
support for CH347T JTAG mode (3)

aystarik:
checked with LA only


trabucayre:
Have you link to a compatbile board and firmware to use ?
Also doc must be updated.

aystarik:
https://github.com/wuxx/USB-HS-Bridge -- here is the board I work with. chip does not require any additional firmware.

aystarik:
checked with real JTAG, it breaks at the usleep() points.

trabucayre:
I have ordered one board but with more or less one month for delivery... I will try to see how to fix/help with that.

aystarik:
great!

PR #332, opened 2023-04-17
Add JTAG to SPI Support on Spartan-6 lx9 CSG324 package.

CreatingNull:
Hi Again,

This PR just adds XC6LX9 JTAG to SPI support for the CSG324 package using the existing `constr_xc6s_csg324.ucf`. I've confirmed the binary works as expected on this package.


trabucayre:
Applied! Thanks!

PR #334, opened 2023-04-17
Allow VID:PID override for JLink probe

viktor-pd:
Add VID:PID override ability for JLink probes. Makes possible to use original Segger JLink PLUS probe which reports VID 0x1366  and PID 0x0101. 
Checked and worked fine programming GOWIN G1WN fpga chip.


trabucayre:
Applied! Thanks!

PR #335, opened 2023-04-19
Add libgpiod v2 api support.

cbalint13:
This PR adds libgpiod >= 2 having new v2 API.

* APIv2 presentation: https://lpc.events/event/16/contributions/1247
* APIv2 docs: https://libgpiod.readthedocs.io/en/latest/index.html

---

* I've tested it with a xc7a200t <--> rk3399 sbc board:

```
# rpm -q libgpiod
libgpiod-2.0.1-1.fc38.aarch64

# ./openFPGALoader -c libgpiod -d /dev/gpiochip1 --pins 3:4:0:1 --detect --verbose-level 99
libgpiod jtag bitbang driver, dev=/dev/gpiochip1, tck_pin=0, tms_pin=1, tdi_pin=3, tdo_pin=4
Raw IDCODE:
- 0 -> 0x13636093
- 1 -> 0xffffffff
- 2 -> 0xffffffff
- 3 -> 0xffffffff
- 4 -> 0xffffffff
found 1 devices
index 0:
	idcode 0x3636093
	manufacturer xilinx
	family artix a7 200t
	model  xc7a200
	irlength 6

# openFPGALoader -c libgpiod -d /dev/gpiochip1 --pins 3:4:0:1 -b acornCle215 \
                                    --verbose-level 99 --dump-flash --file-type bit --file-size 32000000 \
                                    --bitstream dump.bit
libgpiod jtag bitbang driver, dev=/dev/gpiochip1, tck_pin=0, tms_pin=1, tdi_pin=3, tdo_pin=4
Raw IDCODE:
- 0 -> 0x13636093
- 1 -> 0xffffffff
- 2 -> 0xffffffff
- 3 -> 0xffffffff
- 4 -> 0xffffffff
found 1 devices
index 0:
	idcode 0x3636093
	manufacturer xilinx
	family artix a7 200t
	model  xc7a200
	irlength 6
File type : bit
use: /usr/share/openFPGALoader/spiOverJtag_xc7a200tsbg484.bit.gz
load program
Flash SRAM: [=                                                 ] 1.00%^C
```

--

Cc @niklasekstrom , @trabucayre please help with the review.

Thank you,
~Cristian.


src/libgpiodJtagBitbang.cpp
```diff
@@ -57,7 +57,11 @@ LibgpiodJtagBitbang::LibgpiodJtagBitbang(
 	}
 
 	/* Validate pins */
+#ifndef GPIOD_APIV2
 	int pins[] = {_tck_pin, _tms_pin, _tdi_pin, _tdo_pin};
```
trabucayre, line 61: It's possible to also add `const` here?

src/libgpiodJtagBitbang.cpp
```diff
@@ -78,10 +82,52 @@ LibgpiodJtagBitbang::LibgpiodJtagBitbang(
 		throw std::runtime_error("Unable to open gpio chip\n");
 	}
 
+#ifndef GPIOD_APIV2
 	_tdo_line = get_line(_tdo_pin, 0, GPIOD_LINE_REQUEST_DIRECTION_INPUT);
 	_tdi_line = get_line(_tdi_pin, 0, GPIOD_LINE_REQUEST_DIRECTION_OUTPUT);
 	_tck_line = get_line(_tck_pin, 0, GPIOD_LINE_REQUEST_DIRECTION_OUTPUT);
 	_tms_line = get_line(_tms_pin, 1, GPIOD_LINE_REQUEST_DIRECTION_OUTPUT);
+#else
```
trabucayre, line 90: all this part: maybe introducing a private function to use `get_line` or the sequence per pin? This will limited this section of code. If it's possible could you try to respect 80chars/line?

trabucayre:
LGTM. just small nickpick.

niklasekstrom:
It looks good to me as well. I should say I haven't tested it, but from what I can see it should work the same as before.

One thing I reacted to is the negated logic when writing:
```c
#ifndef GPIOD_APIV2
// Old V1 code
#else
// New V2 code
#endif
```

I would find it more intuitive if the condition was positive (not negated):
```c
#ifdef GPIOD_APIV2
// New V2 code
#else
// Old V1 code
#endif
```

But this is a minor detail and perhaps a personal preference thing.

trabucayre:
@niklasekstrom I'm agree for `ifndef` vs `ifdef`, but I have also though it's minor aspect and maybe modified after. The most important thing, in my mind, is to keep code as much as possible easy to maintain.

src/libgpiodJtagBitbang.cpp
cbalint13: > all this part: maybe introducing a private function to use `get_line` or the sequence per pin? 

* Not a straight forward one, **the new()** objects needs some ```_release()``` afterward.
* Maybe if we store into some iterable array these objects by ```get_line()``` and release them afterward.

Not sure if worth it for the four signal lines.

> This will limited this section of code. If it's possible could you try to respect 80chars/line?

* Wasn't sure what linter is used, still there are lines (not touched by me) having > 100 chars.

I've done splitting.

src/libgpiodJtagBitbang.cpp
cbalint13: Done.

src/libgpiodJtagBitbang.cpp
trabucayre: I use `cpplint` but here, in fact, it's because here one line takes two lines :)
For multiple objects for the same pin maybe a small struct is a solution (my first idea was a preprocessor macros but it's a bit too C compliant :) )
something like
```
typedef struct  {
    gpiod_request_config *req_cfg;
    gpiod_line_settings *settings;
    gpiod_line_settings *line_cfg;
    gpiod_line_request *request;
} gpiod_pins;
```
It also simply destructor

cbalint13:
> @niklasekstrom I'm agree for `ifndef` vs `ifdef`, but I have also though it's minor aspect and maybe modified after. The most important thing, in my mind, is to keep code as much as possible easy to maintain.

* I've changed now to ```ifdef``` logic (it was inherited from a debug logic).
* Here is a [build {el8, el9, fc37, fc38, fc39} x {x86_64, ppc64le, aarch64}](https://copr.fedorainfracloud.org/coprs/rezso/HDL/build/5806842/) with latest patch.

I'll wait to finish those and be back to confirm a real FPGA test.

trabucayre:
Thanks!
For init/release I can do that myself in a future commit. It's not really mandatory aspect: just something to reduce code and simplify evolutions.

cbalint13:
> I'll wait to finish those and be back to confirm a real FPGA test.

* The updated ```ifdef``` works:

```
# openFPGALoader -c libgpiod -d /dev/gpiochip1 --pins 3:4:0:1 --detect --verbose-level 99
libgpiod jtag bitbang driver, dev=/dev/gpiochip1, tck_pin=0, tms_pin=1, tdi_pin=3, tdo_pin=4
Raw IDCODE:
- 0 -> 0x13636093
- 1 -> 0xffffffff
- 2 -> 0xffffffff
- 3 -> 0xffffffff
- 4 -> 0xffffffff
found 1 devices
index 0:
	idcode 0x3636093
	manufacturer xilinx
	family artix a7 200t
	model  xc7a200
	irlength 6
```
* With older libgpiod it also compiles fine:
```
-- The CXX compiler identification is GNU 12.2.1
-- Detecting CXX compiler ABI info
{...}
-- Checking for module 'libgpiod'
--   Found libgpiod, version 1.6.4
libgpiod v1 support enabled
```

As final note, the speed remains the same as v1 API (slow, typical for bitbang mode).

trabucayre:
Yes bitbang mode is usual slow.
Unfortunately most faster transport remain FTDI. But I have planned to see if it's possible to improve some code's sections.
I have to have a closer look at v2 specifications/implementation

trabucayre:
Applied. Thanks!

PR #337, opened 2023-04-24
Add support for CH347T in JTAG(3) mode

aystarik:
Completes "detect" without errors.


src/part.hpp
```diff
@@ -181,6 +181,8 @@ static std::map <uint32_t, misc_device> misc_dev_list = {
 	{0x4ba00477, {"ARM cortex A9",         4}},
 	{0x5ba00477, {"ARM cortex A53",        4}},
 	{0xfffffffe, {"ZynqMP dummy device",   12}},
+	{0x1000563d, {"GD32VF103", 5}},
+	{0x790007a3, {"GD32VF103", 5}},
```
trabucayre, line 185: Why this definition? it's because you have a board with a jtag chain containing GiGaDevice + FPGA ?

src/ch347jtag.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <libusb.h>
+
+#include "jtagInterface.hpp"
+
+/*!
+ * \file CH347Jtag.hpp
+ * \class CH347Jtag
+ * \brief concrete class between jtag implementation and FTDI capable bitbang mode
+ * \author Gwenhael Goavec-Merou
```
trabucayre, line 15: I'm not written this class :)

src/ch347jtag.cpp
```diff
@@ -0,0 +1,350 @@
...
+	if ((rlen && rtrans->status != LIBUSB_TRANSFER_COMPLETED) ||
+		 wtrans->status != LIBUSB_TRANSFER_COMPLETED) {
+		return LIBUSB_ERROR_IO;
+	}
+	return 0;
+}
+
+int CH347Jtag::setClk() {
+	uint8_t *ptr = obuf;
+	memset(ptr, 0, 16);
```
trabucayre, line 120: `ptr` is never used: maybe using directly `obuf`?

trabucayre:
I have received my ch347 board and tried to program a Xilinx device but it fails.
I haven't found any information about protocol: could you point me reference? Thanks.

trabucayre:
Ok. Its confirm my assumption :)
I suppose I have to do the same thing to see with writeTDI fails.

aystarik:
I've found that Ghidra was inaccurate, and BinaryNinja made more sense.

trabucayre:
Ok.
One alternate/complementary approach is to use wireshark to dump USB traffic.
BinaryNinja isn't an open source tool no ?

aystarik:
BinaryNinja isn't an open source tool, but their "try" binary works, and https://dogbolt.org/ works with it too.
I've used wireshark -- do not see anything new. The problem is their demo jtag is going to idle state after each dr-shift, so the problem with back-to-back dr-shift is not reproduced...

aystarik:
https://github.com/darioia/ch347_openocd-0.12.0-rc3/blob/master/src/jtag/drivers/ch347_jtag.c -- there is also this code, and there they send last byte as bits regardless of need to switch tms bit, m/b that is the solution...

trabucayre:
If with this approach it's works I think, at least in a first time it's the solution.

aystarik:
tried all the combinations -- nothing works. start to think wch don't have it working either...

trabucayre:
I have to restart my tests. It's a bit weird if this device can't be used...

aystarik:
My doubt about WCH is based on the fact that my setup returns same amount of bytes in D2 case as was sent, or 2 per TCK change, but all the code from WCH treats the response as one byte per TCK, so they should fail every basic test on my version of the chip. BTW, does your chip have a version other than 1.40 of mine ?

aystarik:
Received new boards with newer chip, and it behaves quite different. usleep is no longer needed, bits are counted just right... It might work with this chip version as I can see.

PR #341, opened 2023-05-05
Fix off-by-one errors in calculating UFM start page.

cr1901:
Self-explanatory by the comment. I've personally confirmed this fixes at least the 1200HC parts.


trabucayre:
Applied! Thanks!

PR #351, opened 2023-06-27
Relax ID matching for the misc_dev

UweBonnes:
Detect STM32

Example output:
> ./openFPGALoader -c ftdijtag --detect
index 0:
        idcode   0x6411041
        type     STM32 Device
        irlength 4
index 1:
        idcode   0x4ba00477
        type     ADIv5 JTAG-DP port
        irlength 4

### DIFF
--- a/src/jtag.cpp
+++ b/src/jtag.cpp
@@ -200,7 +200,7 @@ int Jtag::detectChain(int max_dev)
 			 * GateMate and Efinix Trion T4/T8 devices
 			 */
 			if (tmp != 0x20000001)
-				found = search_and_insert_device_with_idcode(tmp & 0x0fffffff);
+				found = search_and_insert_device_with_idcode(tmp);
 			if (!found) /* if masked not found -> search for full */
 				found = search_and_insert_device_with_idcode(tmp);
 
@@ -231,7 +231,7 @@ bool Jtag::search_and_insert_device_with_idcode(uint32_t idcode)
 	if (dev != fpga_list.end())
 		irlength = dev->second.irlength;
 	if (irlength == -1) {
-		auto misc = misc_dev_list.find(idcode);
+		auto misc = misc_dev_list.find(idcode & MISC_DEV_MASK);
 		if (misc != misc_dev_list.end())
 			irlength = misc->second.irlength;
 	}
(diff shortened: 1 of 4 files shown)


src/part.hpp
```diff
@@ -177,10 +177,15 @@ typedef struct {
...
+ * E.g Atmel AVR devices would need a mask 0x0f0003ffU
+ */
+#define MISC_DEV_MASK 0x0ff003ffU
 static std::map <uint32_t, misc_device> misc_dev_list = {
-	{0x4ba00477, {"ARM cortex A9",         4}},
-	{0x5ba00477, {"ARM cortex A53",        4}},
-	{0xfffffffe, {"ZynqMP dummy device",   12}},
+	{0x0ba00077, {"ADIv5 JTAG-DP port",    4}},
+	{0x06400041, {"STM32 Device",          5}},
+	{0x0ff00ffe, {"ZynqMP dummy device",   12}},
```
trabucayre, line 188: The idea here is to provides the full idcode to display exact device (ARM cortex A9 and A53 differs only by upper nibble).
With this modification openFPGALoader will no more able to deal with zynq7000 & zynqMP

src/jtag.cpp
```diff
@@ -200,7 +200,7 @@ int Jtag::detectChain(int max_dev)
 			 * GateMate and Efinix Trion T4/T8 devices
 			 */
 			if (tmp != 0x20000001)
-				found = search_and_insert_device_with_idcode(tmp & 0x0fffffff);
+				found = search_and_insert_device_with_idcode(tmp);
 			if (!found) /* if masked not found -> search for full */
 				found = search_and_insert_device_with_idcode(tmp);
```
trabucayre, line 206: I'm not really happy by the original code because some devices (lattice) with high nibble equal to 0 matches wrong version.
I think both mask & unmask search must be switched, but I have to check to be sure to not introduces a regression.
But If you remove mask most of the devices will never been found because in the idcode list most of devices have highest nibble masked to match all revisions, only a few number of them have full idcode.

UweBonnes:
What about adding a per-vendor  handling after looking for the exact match and before the misc match used to long?

trabucayre:
I have planned to convert the one big list to multiple list (one per vendor) yes.
But not yet made this modification

PR #353, opened 2023-07-10
Tell USB bus and address when ftdi open fails.

inkdot7:
Idea is to give user some more hints when open failed.

### DIFF
--- a/src/ftdipp_mpsse.cpp
+++ b/src/ftdipp_mpsse.cpp
@@ -156,8 +156,18 @@ void FTDIpp_MPSSE::open_device(const std::string &serial, unsigned int baudrate)
 		ret = ftdi_usb_open_bus_addr(_ftdi, _bus, _addr);
 #endif
 	if (ret < 0) {
-		fprintf(stderr, "unable to open ftdi device: %d (%s)\n",
-			ret, ftdi_get_error_string(_ftdi));
+		char description[256];
+		if (_bus == 0 || _addr == 0)
+			snprintf(description, sizeof(description), "");
+		else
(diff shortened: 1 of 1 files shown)


trabucayre:
It's a good idea, but this piece of code is used when `_bus` and `_addr` are provided by user, but also when both aren't. So, in latter case this may introduce difficulties for user.
It's maybe required here to duplicate message to adapt content.

inkdot7:
I think I also were looking for that info when `_bus` and `_addr` had been set automatically.  The extra info might make it easier to figure out what went wrong also in the automatic case.

inkdot7:
Though, my original problem then turned out to be #354, so when giving them explicitly.

How about changing the message from ` (USB bus %d addr %d)` to ` (at USB bus %d addr %d)`?  That would more indicate what failed to open.

trabucayre:
`_bus` and `_addr` have a value != 0 only when:
- user provides both informations `--busdev-num`
- user select a device `--device /dev/ttyUSBx`

So I think it's required to replace the common message by two according to the presence/absence of these informations.

inkdot7:
Oh, now I see.  Yes, with just 0 and 0, it would be more confusing.

I tried to make some formatting that mimicked how the device is opened and noticed that when nothing is specified on the command line, most values (vid, pid, serial, index) are all 0.  So printing them instead would not be helpful.

Though what I made now looks rather clumsy.

Perhaps it is better if, at higher verbosity level, it tells what it tries to open and how.

trabucayre:
Applied thanks @inkdot7

Issue #354, opened 2023-07-10
hex/decimal inconsistency between --scan-usb and --bus-addr


inkdot7:
When using `./openFPGALoader --scan-usb` the bus and address (first numbers) are printed as decimal digits:
```
found 12 USB device
Bus device vid:pid       probe type      manufacturer serial               product
005 010    0x0403:0x6010 FTDI2232        Digilent     210319B4A3B5         Digilent USB Device
```
This is consistent with e.g. `lsusb` which gives:
```
Bus 005 Device 010: ID 0403:6010 Future Technology Devices International, Ltd FT2232C/D/H Dual UART/FIFO IC
Bus 005 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 009 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub
Bus 008 Device 001: ID 1d6b:0001 Linux Foundation 1.1 root hub
...
```
However, using those numbers directly with `--bus-addr` fails if given `5:10`.  It needs to be given the numbers in hexadecimal, `5:a`:
```
root@xcz:~# ~htj/sw/openFPGALoader/build/openFPGALoader --busdev-num 5:010 -b arty ~htj/my_proj.bit
unable to open ftdi device: -3 (device not found) (USB bus 5 addr 16)
JTAG init failed with: unable to open ftdi device
root@xcz:~# ~htj/sw/openFPGALoader/build/openFPGALoader --busdev-num 5:0a -b arty ~htj/my_proj.bit
Jtag frequency : requested 10.00MHz  -> real 10.00MHz 
Open file DONE
Parse file DONE
load program
Flash SRAM: [==================================================] 100.00%
Done
```
Since consistency with `lsusb` is good, if I were to choose, I'd suggest that the argument for `--bus-addr` as default are interpreted as decimal, but that the user can give a `0x` prefix.  That may be bad though, as a `0` prefix then implies octal notation.

inkdot7:
Resolved by #358. @trabucayre  thanks for applying!

PR #358, opened 2023-07-29
Parse USB bus:dev as decimal integers, to be consistent with `--scan-usb`.

inkdot7:
Do not accept auto-detected base 0 (e.g. prefix '0x' for hex), since '0' prefix gives an octal interpretation, and the user is likely to copy values from `--scan-usb` or `lsusb` that could then be misinterpreted, e.g. 005:010.

See description in #354.


trabucayre:
Applied. Thanks @inkdot7

PR #369, opened 2023-08-29
Gowin internal flash works

aystarik:
Re-wrote GOWIN FLASH algorithms so now they seem to work with both authentic ft232 (known to Gowin Programmer) and BL702 emulation on Nano1k board.


trabucayre:
Thanks for this PR.
Could you first split this PR into 2 one for the 3 first commits (unrelated to gowin algo)? This as for goal to focusing here only on gowin update/improvement.
Thanks!

PR #370, opened 2023-08-31
Generic refactoring of the code

aystarik:
Replace std::string for data buffer with the std::vector<uint8_t> as more suitable
Make pointers to its data const so not to screw container itself
Use JTAG state enum instead of integer to better convey sense to compiler
IDCODE is a hex number, so there is no profit in having it signed


trabucayre:
Applied. Thanks @aystarik

PR #371, opened 2023-08-31
GOWIN algorithms re-write

aystarik:
Checked working program SRAM, internal FLASH on Tang Nano1k and Tang Nano 9k with three JTAG cables: original FT232, on-board BL702 emulating FT2232 and CH347 (optimized in this PR).

### DIFF
--- a/src/gowin.cpp
+++ b/src/gowin.cpp
@@ -15,7 +15,6 @@
 
 #include "jtag.hpp"
 #include "gowin.hpp"
-#include "progressBar.hpp"
 #include "display.hpp"
 #include "fsparser.hpp"
 #include "rawParser.hpp"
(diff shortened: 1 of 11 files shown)


src/gowin.cpp
```diff
@@ -745,10 +634,6 @@ bool Gowin::eraseSRAM()
  * but all byte must be bit reversal...
  */
 
-#define spi_gowin_write(_wr, _rd, _len) do { \
-	_jtag->shiftDR(_wr, _rd, _len); \
-	_jtag->toggleClk(6); } while (0)
-
```
trabucayre, line 751: To avoid code duplication (for ease of maintenance/improvement) I prefer keeping this macro. After preprocessor pass code will look  similar to your.

aystarik: It's better to make it a function then, "inline" was invented for this same purpose... Would you like what I update the patch?

PR #374, opened 2023-09-06
update JTAG state machine

aystarik:


trabucayre:
Hi,
I am agree `UPDATE_XR` have the same behaviour and next state are same, but for readability/coherency with state machine I prefer keeping both `UPDATE_XR`.

aystarik:
OK, fair point. What do you think about pre-computing all the TMS transitions beforehand?

trabucayre:
Why not. It's true once next state is known its possible to use an hardcoded two dimensional array to obtain the new value for TMS.

trabucayre:
Looks good.
I have to do a closer look before applying.
Thanks!

PR #379, opened 2023-09-18
JTAG state machine

aystarik:


trabucayre:
Hi,
Do you see any objection I apply some commits manually. There is straigforward ones
- `ch347` and `GD32` are unrelated to the rest and may be applied directly
- it's also true for d03ad1735f2ac0403c1478dcb0d4e503ac391f96 and a45a9e522567b65787e2d95a5472cb99ce770ec0 seems a good improvement.

It will be more easy to have a closer looker to others a bit more tricky.

Thanks!

aystarik:
Hi, not at all. Feel free to cherry-pick however you like. We are not writing on stone tablets :)

aystarik:
Hi, you seem to be hesitant about dropping div_by_5 from MPSSE driver, could you please elaborate on why it was used in first place ? Without it, clock can be from 30 MHz down to 460 Hz, according to AN135 fron FTDI. With it clock can be less than 100 Hz, and we have much larger gaps in frequencies of MHz range, e.g. 2.5 MHz is not possible, only 2MHz and 3MHz are allowed. Do you see range 90 - 460 Hz more important than 1 - 6 Mhz range?

src/gowin.cpp
```diff
@@ -745,10 +668,6 @@ bool Gowin::eraseSRAM()
  * but all byte must be bit reversal...
  */
 
-#define spi_gowin_write(_wr, _rd, _len) do { \
-	_jtag->shiftDR(_wr, _rd, _len); \
-	_jtag->toggleClk(6); } while (0)
-
```
trabucayre, line 751: As mentioned previously I prefer a small, inline, function to avoid too many copy of the same code sequence.

src/gowin.cpp
```diff
@@ -196,307 +199,239 @@ Gowin::~Gowin()
...
+		checkCRC();
 	if (_verbose)
-		displayReadReg(readStatusReg());
+		displayReadReg("after program sram", readStatusReg());
 }
 
-void Gowin::checkCRC()
+void Gowin::program(unsigned int offset, bool unprotect_flash)
 {
-	if (skip_checksum)
```
trabucayre, line 347: It make more sense to keep this (and to let everything about CRC here.

aystarik: it hides the fact that no checksum is checked...

src/gowin.cpp
aystarik: yes, at some point I had a patch with this function added back, must have fell through...

src/gowin.cpp
```diff
@@ -196,307 +199,239 @@ Gowin::~Gowin()
...
-		 * user code register contains checksum or
-		 * user_code when:
-		 * set_option -user_code
-		 * is used: try to compare with this value
-		 */
-		if (hdr.empty()) {
-			is_match = false;
-			snprintf(mess, 256, "Read: 0x%08x checksum: 0x%04x\n",
-				status, checksum);
+		if (!hdr.empty()) {
```
trabucayre, line 375: Why `!hdr.empty()`?
This mean else, where hdr is used is executed only when hdr is empty

aystarik: Right, could be cleaned up more... refreshed

src/gowin.cpp
```diff
@@ -196,307 +199,237 @@ Gowin::~Gowin()
...
-		 */
+	/* no match:
+	 * user code register contains checksum or
+	 * user_code when set_option -user_code
+	 * is used, try to compare with this value
+	 */
+	char mess[256];
+	try {
+		string hdr = _fs->getHeaderVal("checkSum");
 		if (hdr.empty()) {
```
trabucayre, line 374: must be `!hdr.empty()` 
- if empty -> can't compare so fail
- if !empty -> to compare to know ko/ok

src/gowin.cpp
```diff
@@ -196,307 +199,237 @@ Gowin::~Gowin()
...
-					"Read 0x%08x (checksum: 0x%08x, user_code: 0x%08x)\n",
-					status, checksum, user_code);
+			if (ucode != user_code) {
+				snprintf(mess, 256, "Read 0x%08x (checksum: 0x%08x, user_code: 0x%08x)\n",
+					 ucode, checksum, user_code);
+				printError("CRC check : FAIL");
+				printError(mess);
 			}
+			printSuccess("CRC check: Success");
 		}
```
trabucayre, line 383: maybe:
```
if (ucode == user_code) {
    printSuccess("CRC check: Success");
    return;
}
```

So if no match: message in lines 385-387 is displayed and this avoid to have twice error handling.

Also, in your code if fail, since there is not `return` the success message is always displayed

aystarik: right, rewrote it

src/gowin.cpp
```diff
@@ -517,220 +446,213 @@ bool Gowin::pollFlag(uint32_t mask, uint32_t value)
...
+	_jtag->shiftDR((uint8_t *)&checksum, NULL, 32);
+	send_command(0x08);
+
+	send_command(CONFIG_DISABLE); // config disable
+	send_command(NOOP); // noop
+	
+	if (_verbose)
+		displayReadReg("after write sram", readStatusReg());
+	if (readStatusReg() & STATUS_DONE_FINAL) {
+		printSuccess("DONE");
```
trabucayre, line 552: please keep progressBar (I know it's a slow xfer but it's more visual for user)

src/gowin.cpp
```diff
@@ -517,220 +446,228 @@ bool Gowin::pollFlag(uint32_t mask, uint32_t value)
...
+
+	if (_verbose)
+		displayReadReg("after erase sram", readStatusReg());
+	if (readStatusReg() & STATUS_DONE_FINAL) {
 		printError("FAIL");
 		return false;
+	} else {
+		printSuccess("DONE");
+		return true;
 	}
```
trabucayre, line 663: Why fail when flag is set?
Maybe also keeping test for `STATUS_MEMORY_ERASE` to check if this step is fine.

aystarik: Take a look at UG290 p.38, figure 6-15. DoneFinal==0 is success for Erase SRAM and Erase FLASH
My understanding is that DoneFinal is FPGA has loaded config in SRAM. There is no flag for success of FLASH erase as such, so they issue load after erase and if FPGA stays clean, FLASH have been erased.

trabucayre:
Sorry for the big long delay.
I'm okay, I wish just to take time to do some test with real hardware. I have experimented, by the past, some issue with gowin devices (mainly littlebee) when something goes wrong.

trabucayre:
I have tried tomorrow with a gw1n1: SRAM is working but FLASH no. I have to compare original code with modified one to see why.

aystarik:
Did erase flash happen? Major difference in erase flash -- I've made "shift-dr 32 bit" only for GW1N-4 as written in documentation, it helped with erase of GW1N-9, but may be GW1N-1 still prefers to have it. In programming main difference with original core is that I do not add blocks before the bitstream, but modify 4 bytes with auto-load pattern in bitstream itself. I've checked with LA, and gowin programmer does exactly that. Actually, may be, you'll get some insight to what gowin programmer intentions are by using programmer_cli to generate svf file -- it can do that for flash programming, and it somewhat easier to read than the LA log.

trabucayre:
Hi,
I have created a [branch](https://github.com/trabucayre/openFPGALoader/tree/rework_gowin) with a mix between your PR and the master branch. There is a few modifications (mainly to use disableCfg/enableCfg and writeFLASH is original one but I have to compare and merge). This version works with my gw1n1 device (and I have to test with all FPGAs I have).

Using SVF is a good way but has some drawback: there is no loop so a wait for a ''timeout'' duration is applied before simply reading status with a delay more or less dependent to the device and not always true for all. When using ILA or sniffing  USB traffic you have sometime access to a more rich information. So a mix between these two approach is sometime good.

trabucayre:
In fact I have observed the 5x32 dummy bits by reading USB trafic, maybe this part is no more true and was a gowin programmer's bug.
I have to test by simply adding pattern at the beginning of the flash

aystarik:
in my [devel branch](https://github.com/aystarik/openFPGALoader/tree/devel ) i have made a small exec which only reads the fs file -- it may be of use to compare the data from LA and from the file

aystarik:
BTW, I can't find TN653 listed in gowin docs site, it might have been superseded by UG290

trabucayre:
You can find this AN [here](http://cdn.gowinsemi.com.cn/TN653E.pdf)

I have tried by removing dummy byte and it works.
When I have (2019) started to integrate gowin FPGAs I have compared fs file with USB packets content and seen this sequence (pattern + dummy) before configuration data. Maybe it's a bug somewhere, or a solution to obfuscate trafic...
The question here is to know if this pattern must be BEFORE fs or if the firt 4B must be overriden with this sequence...

aystarik:
I've compared whole file data with output from LA (whole session) and patterns replace first 4 bytes in the data stream. Everything else is the same between file and LA stream

aystarik:
No, I have no problem _downloading_ the 2019 version of this file by link -- programmer manual still has one after the ug290 link. I've pointed out that it is not possible to find it in the _list_ of documentation.

trabucayre:
Ah ok. sorry for the mistake. I keep this TN because some informations was lost on UG290 (in fact I play with both app notes to find required informations).
Maybe I have to add a comment with the direct link to this file.

aystarik:
Yes, documentation is worst chinese style -- text does not explain anything, figures contradict text, and  there are several documents which contradict each other and the output of their own program... BTW, have you noticed that it is possible to write auto-boot pattern over read-back pattern without flash erase -- it only needs to clear some more bits.
'''
read_pattern = 0xf0f00f0f|ab_pattern
'''

trabucayre:
Hi.
I have tested code with more or less all Gowin I have and updated my branch.
If you're agree with modifications I apply this patch.

For readback vs autoboot: it's may interesting to add this feature to verify code after write (or to force autoboot pattern after write).

Thanks.

aystarik:
Hi, looks good, thanks!

trabucayre:
I have applied gowin commit. Now it remains jtag rework part.

Thanks

aystarik:
Do you need any help with jtag parts?

trabucayre:
Not now. I have to review and to find the best compromise between speed and pedagogy.

aystarik:
> ... to find the best compromise between speed and pedagogy.
Hi, could you please describe "compromise between speed and pedagogy" in more detail? I struggle to even put these words in the same sentence...

trabucayre:
Exactly.
In one hand there is orignal code with detail on state transitions. But this code is slow since same transition are computed all the time.
In another hand there is your approach where all transition are precomputed (at compile time ?) and automatically faster. But it's less visual/pedagogic.

aystarik:
>  But it's less visual/pedagogic.

My thinking was that printing of transitions from tms bits is enough? It is the commented out code in checking the correctness of tables in the beginning of the file? May be to use it if verbose was asked for?

aystarik:
dive() -- this is recursive depth first search of the shortest path on the directed graph of the JTAG state machine. create_paths() calls it for every combination of states thus creating square array of paths. create_paths is called from Jtag constructor once. print_path() can then pretty-print computed TMS sequence.

PR #381, opened 2023-09-23
Add support for Xilinx xczu17eg. It's tested on xczu17eg board with D…

openhardwarefan:
…igilent HS3 at 30Mbps.


trabucayre:
Applied. Thanks @openhardwarefan

PR #390, opened 2023-10-12
Lattice programming fix and boards added

bg-gsl:
The edits to the Lattice programming procedure are in 2 areas:
1. it can now handle the status register for the Nexus family, which is 64 bits wide, compared to the 32 bits of other families. In the upper 32 bits (34-37), there are the "BSE Error 1 Code for previous bitstream execution" bits, which tell us if there were issues in the current fpga bitstream. This is useful when e.g. there's no bitstream available in the SPI Flash and it's tied to the next point.
2. When the FPGA is in a "error state" (some bits in "BSE Error 1 Code for previous bitstream execution" are 1), a plain load of a new bitstream will not clear the "error state". An example is that the INITN led on the board stays lighted up after programming is done, even though the new bitstream seems to work fine. To fix this issue, we need to send a LSC_REFRESH command (which needs a small waiting time after, as otherwise the BUSY flag doesn't get cleared) and then reset the configuration logic. This is not needed if the FPGA starts in a non "error state" and the usual programming procedure works fine.

Lattice documents: FPGA-TN-02099-1.4

Ideas for such fixes came from analyzing the svf file generated from a bitstream for CertusNX using Radiant Programmer. Here [svf_demo.txt](https://github.com/trabucayre/openFPGALoader/files/12878499/svf_demo.txt) you can find the txt version of such file with some fields "censored" (including the bitstream data). In this svf file, all instructions described above are executed anyway.

Edits were tested on Crosslink Eval board, Certus Versa Eval board, CertusPro Eval board and CertusPro Versa Eval board, all of Nexus family. No test was done on other families, as we don't have access to those fpgas.

Boards added are: Certus Versa Eval board, CertusPro Eval board and CertusPro Versa Eval board.
To add the CertusPro boards, we added the part number for the CertusPro fpga as well.


trabucayre:
LGTM!
Thanks for this work!
Just some nitpicking, but if you prefer I can do it by myself after merge.

Thanks!

bg-gsl:
Hej!
It's fine for me to proceed in the way suits you best :)
As a reminder: I've realised I didn't add anything to the documentation when it comes to  the supported boards.

trabucayre:
Applied
I have enough informations to update doc part
Thanks @bg-gsl

trabucayre:
I have [updated](https://github.com/trabucayre/openFPGALoader/commit/f964c3364f230b948c53dfe485479c4d2612c2d9) doc with informations you have provided.
Thanks again

PR #395, opened 2023-10-20
Add ft4232hp, gr740-mini, mappings of spi flash for Lattice boards, small fixes for Lattice programming procedure

bg-gsl:
Hello!

Here are the main changes:
- Add GR740-mini board (which features a Lattice CertusPro fpga): https://gaisler.com/index.php/products/boards/gr740-mini
- Add FTDI FT4232HP cable (+ lines mapping): https://ftdichip.com/wp-content/uploads/2023/02/DS_FT4233HP.pdf . This FTDI chip is used on the GR740-mini board.
- Spi flash memories mapping:
    - Add Macronix MX25L51245G: this is used in the CertusPro Versa and GGR740-mini boards
    - Add Micron MT25Q128: this is used in the CertusPro eval and Certus Versa boards. The JEDEC IDs are the same as the N25Q128 (0x20bb18). Differences between the two are reported here: https://media-www.micron.com/-/media/client/global/documents/products/technical-note/nor-flash/tn2501_migrating_n25q_to_mt25ql.pdf
    - Distinguish between N25Q128 1.8V and N25Q128 3V, as they have different JEDEC IDs: 0x20bb18 vs 0x20ba18 respectively
    - Links to datasheets are added in the comments.
- Fixes in Lattice programming procedure:
    - The main thing is that the Nexus family has a Bscan register 45.25 bytes wide and therefore the previous PRELOAD operation (set to shift 26 bytes) didn't clear/reset the full register: we could notice indeed that programming certain fpgas/boards required 2 consecutive programmings to get things to run correctly.
    - REG_STATUS_PRV_CNF_CHK_MASK: the "Configuration Status Check of previous bitstrem" starts at bit 34 (my previous mistake)
    - fix the warning "left shift count >= width of type" about REG_STATUS_PRV_CNF_CHK_MASK initialization

The final version of openFPGALoader have been tested with all the Lattice Nexus boards that the project supports and the GR740-mini.

Only one thing to highlight/warn about: The programming of the SPI Flash is not 100% reliable. While on some boards things go well (Certus versa and CertusPro Versa), on some (e.g. CertusPro eval) results are mixed (sometimes it works, some other times not). I currently don't have time to dig deep into the SPI flash programming part, sorry.

Let me know if something needs to be changed/removed! :)


trabucayre:
LGTM. Just a few nitpick.
Thanks!

bg-gsl:
Great, thanks!

In case : a colleague reported a possible mistake here: https://github.com/trabucayre/openFPGALoader/blob/master/src/lattice.cpp#L1026
It seems that the type of the input should be changed: `(dev >> 1)` => `( (unsigned int) dev >> 1)`

trabucayre:
Applied thanks @bg-gsl !

PR #399, opened 2023-10-26
Fix lattice bscan nexus in clearSRAM()

bg-gsl:
For NEXUS family fpgas, the Bscan register is 362 bits long or 45.25 bytes => 46 bytes.
This error was already corrected in `program_mem` (to program the sram).

In our case, we were interested in programming the external SPI Flash memory.

Tests have been carried out successfully on Certus-NX Versa, Crosslink-NX Evalualtion, CertusPro-NX Versa and GR740mini.

On CertusPro-NX Evaluation board instead, results were mixed: if the spi flash was not already programmed, then things were going fine. Otherwise, openFPGALoader doesn't manage to read/talk to the SPI Flash memory. We've already had issues with such board when programming the spi flash, so it may be that the sample we have in office is not working well.

### DIFF
--- a/src/lattice.cpp
+++ b/src/lattice.cpp
@@ -719,6 +719,8 @@ bool Lattice::clearSRAM()
 	}
 	wr_rd(PRELOAD_SAMPLE, tx_buf, tx_len, NULL, 0);
 
+	wr_rd(0xFF, NULL, 0, NULL, 0);
+
 	/* ISC Enable 0xC6 */
 	printInfo("Enable configuration: ", false);
 	if (!EnableISC(0x00)) {
--- a/src/lattice.cpp
+++ b/src/lattice.cpp
@@ -729,7 +729,7 @@ bool Lattice::clearSRAM()
 		printSuccess("DONE");
 	}
 
-	if (_fpga_family == MACHXO3D_FAMILY)
+	if (_fpga_family == MACHXO3D_FAMILY || _fpga_family == NEXUS_FAMILY)
 		erase_op = 0x0;
 	else
 		erase_op = FLASH_ERASE_SRAM;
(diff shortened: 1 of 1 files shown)


src/lattice.cpp
```diff
@@ -704,12 +705,19 @@ bool Lattice::clearSRAM()
 {
 	uint32_t erase_op;
 
-	/* preload 0x1C */
-	uint8_t tx_buf[26];
-	memset(tx_buf, 0xff, 26);
-	wr_rd(0x1C, tx_buf, 26, NULL, 0);
-
-	wr_rd(0xFf, NULL, 0, NULL, 0);
```
trabucayre, line 712: This operation isn't required?

trabucayre: more precisely `wr_rd`at line 712

bg-gsl: mmhh, I may have deleted it accidentally. I'll check tomorrow (I created an svf file from Radiant) and test it on a board.

Thanks for pointing it out!

bg-gsl: I've tried with and without and I don't see differences in execution. If you prefer to keep it, then it's ok with me :)

bg-gsl: Hej! :)
Just to avoid misunderstandings: is there something I need to do about this pull request?

trabucayre: Sorry: my fault I have missed to answer.
In fact this code is used by all lattice devices so it's maybe more secure to keep/re-add 0xff instruction.

bg-gsl: I've restored that line :)

trabucayre:
Applied. Thanks @bg-gsl

PR #401, opened 2023-10-28
ALINX AX7101 board.

inkdot7:
Uses same FPGA-board as on the carrier AX7102, so essentially same as #397.


trabucayre:
Applied. Thanks @inkdot7 !

PR #409, opened 2023-12-08
Add pynq-z1 board

timkpaine:
Board is essentially 1-1 with pynq z2, but more expensive (for non academic).


trabucayre:
Applied. Thanks @timkpaine

PR #410, opened 2023-12-12
gatemate: fix configuration byte alignment in jtag chains

pu-cc:
When used in a JTAG chain with multiple devices - and due to the additional bypass cycles - the 8-bit words in the DR register are not always contiguous. This PR sends the corresponding sequence of zeros so that there are always correct multiples of 8.

This allows configuration of individual GateMate FPGAs in a chain (using --index-chain).

### DIFF
--- a/src/colognechip.cpp
+++ b/src/colognechip.cpp
@@ -7,6 +7,7 @@
 #include "colognechip.hpp"
 
 #include <memory>
+#include <string.h>
 
 #define JTAG_CONFIGURE  0x06
 #define JTAG_SPI_BYPASS 0x05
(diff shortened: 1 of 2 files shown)


trabucayre:
LGTM. But before applying: this is something specific to GateMate? openFPGALoader, when more than one device is present, already handle nop before or after and the addition of bits to align data.

pu-cc:
I wouldn't say it's too specific.
The configuration controller only does not recognize a command/byte if it is in a jtag chain with multiple taps as it always reads 8 consecutive bits:

Assume there are 3 taps in the chain and we want to configure the 3rd tap. After the IR has been loaded with the command 0x06 (JTAG_CONFIGURE), openFPGALoader correctly sends two nops for the bypass registers of the first two taps. However, because the DR in the 3rd tap is also shifted by TCK, the resulting command is shifted by 2. This PR checks whether the selected TAP is in a chain and sets the corresponding DR-NOP-shifts to align the command. In this case, there would be 6 additional DR-nop shifts. The first command/byte would then be 0x00 and all following commands are correctly aligned.

trabucayre:
Ok. I understood better.
In fact I see one issue in the code (not this PR but in original).
In line 289, next state is always SHIFT_DR and a set_state is used to move to RUN_TEST_IDLE.
shiftDR is able to push a sequence of '0' after the last bit for the last packet when next_state is set to a state different to SHIFT_DR.
Could you try this change? I know people using openFPGALoader with > 40 FPGAs in the chain without any issue.
If this modification isn't working I will apply your PR.

Something like:
```
    ProgressBar progress("Load SRAM via JTAG", length, 50, _quiet);

    Jtag::tapState_t next_state = Jtag::SHIFT_DR;
    for (int i = 0; i < length; i += size) {
        if (length < i + size) {
            size = length-i;
            next_state = Jtag::RUN_TEST_IDLE;
        }

        for (int ii = 0; ii < size; ii++)
            tmp[ii] = data[i+ii];

        _jtag->shiftDR(tmp, NULL, size*8, next_state);
        progress.display(i);
    }

    progress.done();

    if (_ftdi_jtag) {
        waitCfgDone();
        _ftdi_jtag->gpio_set(_oen_pin);
    }
```

Thanks!

pu-cc:
Thank you! I checked it, but it does not address the problem with the proper alignment. After the IR of the target tap has been latched, it expects the config stream in the DR path. Because of the additional bypass clocks, we still have to correct the alignment. Other manufacturers probably align via some magic sequence right at the beginning of the configuration steam; but gatemate doesn't have anything like that.

Your approach to getting into RUN_TEST_IDLE state after sending the config stream is certainly much better.

trabucayre:
Merged thank!
And thanks  to have checked.

PR #412, opened 2023-12-22
Add WebAssembly support

whitequark:
I got openFPGALoader to work with WebUSB with this patch (tested on OrangeCrab r0.2).

### DIFF
--- a/src/gowin.cpp
+++ b/src/gowin.cpp
@@ -231,7 +231,7 @@ bool Gowin::send_command(uint8_t cmd)
 #include <libkern/OSByteOrder.h>
 #define le32toh(x) OSSwapLittleToHostInt32(x)
 #define htole32(x) OSSwapHostToLittleInt32(x)
-#elif (defined(_WIN16) || defined(_WIN32) || defined(_WIN64)) || defined(__WINDOWS__)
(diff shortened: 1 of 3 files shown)


trabucayre:
Applied! Thank @whitequark

whitequark:
Thanks! Do you think I should be shipping releases of openFPGALoader as a part of [YoWASP toolchain](https://marketplace.visualstudio.com/items?itemName=yowasp.toolchain) or if I should ship the master branch?

trabucayre:
Good question.  I think master branch is better:
1. Current master branch has improvements for nexus devices and support for gowin GW5A so It seems  better in a first time to use it.
2.  more generically I create a new release more or less each 6 month. It a lood delay when improvement are applied.
3. your fixes are on master branch.

But in other hand I try to fix all known issues and fix regression before tagging a commit (this why I not iregular for releases).

whitequark:
It sounds like adding automation that ships the master branch nightly is the way to go here. (Obviously I can change this later, I have scripts tailored for different workflows.)

trabucayre:
If it's possible without to many works it seems good yeah. 
Or I can create a tag by incrementing minor number if you prefer.

whitequark:
Already set up in fact! https://www.npmjs.com/package/@yowasp/openfpgaloader?activeTab=versions

trabucayre:
Great! Thanks a lot for your work!

PR #419, opened 2024-01-18
lattice: Add all MachXO[23] part IDs

sean-anderson-seco:
The version field is the only difference between many parts in the MachXO[23] family, including between different families. Add the version field to all parts, fixing detection of some MachXO3 parts as MachXO2s. The id codes were extracted from the BSDL files on Lattice's website.


trabucayre:
Applied Thanks @sean-anderson-seco !
Removing highest nibble for all devices wasn't a good idea an must be limited to vendor where this is relevant

PR #424, opened 2024-02-01
updata CH347

ZhiyuanYuanNJ:
Added CH347F, which is a new package version of CH347T. 
BUG fix: Sometimes writing bit files may fail.
Added functions: Added VID, PID and other settings, the default selected model is CH347T.


src/ch347jtag.cpp
```diff
@@ -22,14 +22,20 @@
...
+#define GHZ(n) (uint32_t)((n)*UINT32_C(1000000000))
 
 #define CH347JTAG_INTF        2
 #define CH347JTAG_WRITE_EP    0x06
 #define CH347JTAG_READ_EP     0x86
 
 #define CH347JTAG_TIMEOUT     200
-
+unsigned short CH347_VID = CH347JTAG_VID;
+unsigned short CH347_PID = CH347JTAG_PID;
```
trabucayre, line 38: Variables must be private members of the class (ie in .hpp file).
But here vid/pid are only used in constructor method so there is no real needs to store these informations

src/ch347jtag.cpp
```diff
@@ -164,19 +143,41 @@ int CH347Jtag::setClk(const uint8_t &factor) {
...
 {
+	libusb_device** devs;
 	int actual_length = 0;
+	int i = 0;
+	ssize_t cnt;
 	struct libusb_device_descriptor desc;
 	struct libusb_device *dev;
 	int rv;
+    CH347_VID = vid;
+	CH347_PID = pid;
```
trabucayre, line 157: ditto. Not required

src/ch347jtag.cpp
```diff
@@ -164,19 +143,41 @@ int CH347Jtag::setClk(const uint8_t &factor) {
...
+				if (libusb_open(dev, &dev_handle) < 0){
+					printError("libusb init failed");
+					goto err_exit;
+				}
+			}
+		}
+	}else {
+		dev_handle = libusb_open_device_with_vid_pid(usb_ctx, CH347_VID,
+			CH347_PID);
+	}
```
trabucayre, line 180: The code used here to search for a device using bus/dev addr is quite similar to `libusb_open_device_with_vid_pid` implementation. It's maybe possible to merge both to:
1. for each device compare vid/pid
2. if `bus_addr` and  `dev_addr` are provided do comparison too

Something similar to:
```
[...]
if (desc.idVendor != CH347_VID || desc.idProduct != CH347_PID)
    continue;
if (bus_addr != 0 && dev_addr != 0 && (libusb_get_bus_number(dev) != bus_addr || libusb_get_device_address(dev) != dev_addr))
    continue;
```

src/ch347jtag.cpp
```diff
@@ -243,37 +244,33 @@ CH347Jtag::~CH347Jtag()
...
-			break;
-	}
-#endif
-	_clkHZ = sl;
-	if (setClk(i)) {
+    int setClk_index = 0;
+	uint32_t speed_clock[8] = {
+		KHZ(468.75), KHZ(937.5), MHZ(1.875), MHZ(3.75),
+		MHZ(7.5), MHZ(15), MHZ(30), MHZ(60)
+	};
```
trabucayre, line 251: Clock speed/index depends on device (**STANDARD_PACK** and **LARGER_PACK**). This must be keep in mind/used here

src/ch347jtag.cpp
```diff
@@ -243,37 +244,33 @@ CH347Jtag::~CH347Jtag()
...
 int CH347Jtag::writeTMS(const uint8_t *tms, uint32_t len, bool flush_buffer,
 		__attribute__((unused)) const uint8_t tdi)
 {
-	if (get_obuf_length() < (int)(len * 2 + 4)) { // check if there is enough room left
-		flush();
-	}
+	// if (get_obuf_length() < (int)(len * 2 + 4)) { // check if there is enough room left
+	// 	flush();
+	// }
+ 	flush();
```
trabucayre, line 273: Why always flushing?

src/ch347jtag.cpp
```diff
@@ -424,12 +421,12 @@ int CH347Jtag::writeTDI(const uint8_t *tx, uint8_t *rx, uint32_t len, bool end)
...
-		uint8_t *xb = &ibuf[3 + i * 8];
-		for (unsigned j = 0; j < 8; ++j)
-			b |= (xb[j] & 1) << j;
-		*rptr++ = b;
+	for (unsigned i = 0; i < size; ++i) {
+		if (ibuf[3 + i] == 0x01) {
+			*rptr |= (0x01 << i);
+		}else{
+			*rptr &= ~(0x01 << i);
+		}
```
trabucayre, line 429: Here buffer received is in bit mode: one byte received contains the value for on state of TDO.
So each byte must be reconstructed.
Also rptr is never updated so you always write at the same position.
Why replacing original code?

src/ch347jtag.cpp
```diff
@@ -51,17 +57,17 @@ static void LIBUSB_CALL sync_cb(struct libusb_transfer *transfer) {
...
 
-int CH347Jtag::usb_xfer(unsigned wlen, unsigned rlen, unsigned *ract, bool defer) {
+int CH347Jtag::usb_xfer(unsigned wlen, unsigned rlen, unsigned *ract, bool defer) 
+{
+	int actual_length = 0;
 	if (_verbose) {
 		fprintf(stderr, "usb_xfer: deferred: %ld\n", obuf - _obuf);
 	}
-	if (defer && !rlen && obuf - _obuf + wlen < MAX_BUFFER - 12) {
+	if (defer && !rlen && obuf - _obuf > (wlen + 12)) {
```
trabucayre, line 66: obuf may point out of _obuf area if the size (MAX_BUFFER) is not taken into account here

src/ch347jtag.cpp
ZhiyuanYuanNJ: ok, It will be deleted.

src/ch347jtag.cpp
ZhiyuanYuanNJ: Changed.

src/ch347jtag.cpp
ZhiyuanYuanNJ: Changed.

src/ch347jtag.cpp
ZhiyuanYuanNJ: Sorry, I don't understand what you mean STANDARD_PACK.

src/ch347jtag.cpp
ZhiyuanYuanNJ: If flushing is not done in time, will TMS changes possibly be missed in the last step of the process?

src/ch347jtag.cpp
ZhiyuanYuanNJ: This represents the value of each bit on the last byte of TDO, and the value of the rptr binary bit has been changed.

src/ch347jtag.cpp
ZhiyuanYuanNJ: Changed.

src/ch347jtag.cpp
```diff
@@ -164,19 +142,34 @@ int CH347Jtag::setClk(const uint8_t &factor) {
...
+			if (desc.idVendor != vid || desc.idProduct != pid)
+    			continue;
+			if (bus_addr != 0 && dev_addr != 0 && (libusb_get_bus_number(dev) != bus_addr || libusb_get_device_address(dev) != dev_addr))
+    			continue;
+			libusb_open(dev, &dev_handle);
+			break;
+		}
+	}else {
+		dev_handle = libusb_open_device_with_vid_pid(usb_ctx, vid, pid);
+	}
```
trabucayre, line 172: In fact here my idea is to remove `libusb_open_device_with_vid_pid`and the test in l.159.

src/ch347jtag.cpp
trabucayre: seems my previous comment was lost.
https://github.com/YTEC-info/CH347-Softwares/blob/main/Datasheet%20%26%20Manual/WCH-CH347-JTAG-Interface(1.3).pdf
explains according to model and firmware the is two mode.
Each of them as specific frequencies.

src/ch347jtag.cpp
trabucayre: It must be tested, but usually I prefer using a flush just before communication (TDI/TDO). But if its required here: its fine

src/ch347jtag.cpp
trabucayre: `*rptr`always point to the same adress, you have to increment memory position each 8bits.
Also you shift using `i, if `size == 256` the bit will be lost because `*rptr` is a 8bit.
Again I don't see (but maybe I'm wrong) any issue with the original code.

ZhiyuanYuanNJ: The size here will not be greater than 8. If chip returns 10bit data 0b1100111101, then 347 will be packaged as D4 00 01 CF D2 02 00 00 01, so according the original code  **size(here is 2)  / 8**, the last 2 bits will be missed.

src/ch347jtag.cpp
ZhiyuanYuanNJ: You mean like this?
image

src/ch347jtag.cpp
ZhiyuanYuanNJ: I also prefer this. I changed it to always flush() because I am worried about after the last time communication(TDI/TDO), and flush condition is not satisfied at this time, then will the process get stuck?

src/ch347jtag.cpp
ZhiyuanYuanNJ: Changed.

src/ch347jtag.cpp
trabucayre: Yes!
maybe a `break` after `libusb_open` to only select the first when bus_addr and dev_addr are not provided?
Also `libusb_free_device_list(devs, 1)` must be used just after the loop and before `if (!dev_handle) {`

ZhiyuanYuanNJ: You are right, thanks for pointing out the error.
image

trabucayre:
Applied. Thanks @ZhiyuanYuanNJ !

PR #427, opened 2024-02-02
Fix Tang Primer 25K SRAM loading when flash is erased

### LINKED ISSUE #426: Tang Primer 25K SRAM loading fails when flash is erased
openFPGALoader fails to load _Tang Primer 25K_ SRAM when its flash has been erased and the board was powercycled.

Gowin programmer works fine in this case. Monitoring USB packets reveals that Gowin  programmer sends extra commands before SRAM erase + load process if flash was erased beforehands. The commands in the marked part below are not sent when flash contains valid data. This part is probably injected after inspecting Status Register bits signalling invalid flash contents.
The most interesting would probably be the command **0x3F** which I was not able to locate any reference for in Arora V datasheet.

```
TAP reset
CMD 3A        // Config Disable
RD 1 byte     // execution result
CMD 00        // ???
RD 1 byte     // execution result
TAP reset
CMD 41        // Read Status Register
RD 1 byte     // execution result
R/W 7 bytes   // > FFFF FFFF 0000 00 (last byte 3 bits only are actually clocked out)
RD 7 bytes    // < 3862 0276 FFFF 80

==== ERASED FLASH ONLY START ====
CMD 3F        // ??? Undocumented command
RD 1 byte     // execution result
--- wait 100 ms
CMD 02        // NOOP
RD 1 byte     // execution result
CMD 11        // Supposedly Read ID code, but the code is never actually read
RD 1 byte     // execution result
CMD 02        // NOOP
RD 1 byte     // execution result
Clock out [19] 126x 00 byte   // ~ 100 us delay
--- wait 100 ms
CMD 11        // Supposedly Read ID code, but the code is never actually read
RD 1 byte     // execution result
==== ERASED FLASH ONLY END ====

CMD 15        // Config Enable
...
```

jgroman:
Fixes #426 

I am not sure how to properly detect invalid/erased flash on Tang Primer 25K, but the "auto_boot_2nd_fail" status bit seems like a good candidate.

With this modification I am able to load Tang Primer 25K (GW5A-25) SRAM successfully regardless of the flash contents.

### DIFF
--- a/src/gowin.cpp
+++ b/src/gowin.cpp
@@ -883,8 +883,19 @@ void Gowin::sendClkUs(unsigned us)
 bool Gowin::eraseSRAM()
 {
 	printInfo("Erase SRAM ", false);
+	uint32_t status = readStatusReg();
 	if (_verbose)
-		displayReadReg("before erase sram", readStatusReg());
+		displayReadReg("before erase sram", status);
+
+	// If flash is invalid, send extra cmd 0x3F before SRAM erase
+	// This is required on GW5A-25
+	bool auto_boot_2nd_fail = (status & 0x8) >> 3;
+	if ((_idcode == 0x0001281B) && auto_boot_2nd_fail)
+	{
+		disableCfg();
+		send_command(0x3F);
+		send_command(NOOP);
+	}
 
 	if (!enableCfg()) {
 		printError("FAIL");


trabucayre:
Applied! Thanks @jgroman

PR #431, opened 2024-02-12
SVC client fixup error handling

mdavidsaver:
In trying out openFPGALoader with a new (and therefore buggy) XVC server I managed to put openFPGALoader into a tight loop printing `Send instruction failed` when that server crashed.

Somewhat crudely, this PR `throw`s a c++ exception in the places where `ll_write()` errors were being ignored.  These exceptions are not caught, and so `abort()`.  An improvement, although not ideal.  It isn't clear to me how to communicate failure back through `Xilinx::program()`.

Also, add `sendall()` logic as TCP socket `send()` may enqueue some, but not all, of the requested bytes.

### DIFF
--- a/src/xvc_client.cpp
+++ b/src/xvc_client.cpp
@@ -242,6 +242,23 @@ bool XVC_client::open_connection(const string &ip_addr)
 	return true;
 }
 
+static
+int sendall(int sock, const void* raw, size_t cnt, int flags)
+{
+	const char *buf = (const char*)raw;
+	size_t remaining = cnt;
+	while (remaining) {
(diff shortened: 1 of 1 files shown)


trabucayre:
LGTM.
Yes usual way is  to use return with an error code and let caller deal with this value, but it's true in many part this error code isn't evaluated.
Reasons are mainly:
* some cable drivers don't return relevant informations
* coherency issue between them.
* at higher level (vendor level) a correct policy about how to deal with this information must be implemented (stop immediately, recover mode, ...)

trabucayre:
Applied. Thanks @mdavidsaver

Issue #436, opened 2024-02-26
Missing delay after loading spioverjtag?


UweBonnes:
Hello,
first loading the spioverjtag bitfile and then writing the flash with skipping the load step detects the flash, also the flash is not know yet.

> ./openFPGALoader -b vmm3 ../spiOverJtag/spiOverJtag_xc7s50csga324.bit.gz 
empty
Can't read iSerialNumber field from FTDI: considered as empty string
Jtag frequency : requested 6.00MHz   -> real 6.00MHz  
Open file DONE
Parse file DONE
load program
Load SRAM: [==================================================] 100.00%
Done
> ./openFPGALoader --skip-load-bridge -b vmm3 -f /tmp/44MHz.mcs
empty
write to flash
Can't read iSerialNumber field from FTDI: considered as empty string
Jtag frequency : requested 6.00MHz   -> real 6.00MHz  
Open file DONE
Parse file DONE
Skip loading bridge for spiOverjtag
Detail: 
Jedec ID          : 01
memory type       : 60
memory capacity   : 18
flash chip unknown: use basic protection detection
Erasing: [==================================================] 100.00%
Done
Writing: [==================================================] 100.00%
Done

Doing it in one step fails to detect the SPI flash:
> env OPENFPGALOADER_SOJ_DIR=../spiOverJtag/ ./openFPGALoader -b vmm3 -f /tmp/44MHz.mcs
empty
write to flash
Can't read iSerialNumber field from FTDI: considered as empty string
Jtag frequency : requested 6.00MHz   -> real 6.00MHz  
Open file DONE
Parse file DONE
use: ../spiOverJtag//spiOverJtag_xc7s50csga324.bit.gz
load program
Load SRAM: [==================================================] 100.00%
Done
Detail: 
Jedec ID          : ff
memory type       : ff
memory capacity   : ff

If there is a way to detect fpga reboot after loading the fpga sram, does openFPGALoader wait for that event? Or must we add some delay?

trabucayre:
There is a delay after `JTSTART` and I have already tried to flash spartan7 without seeing this type of issue (I have digilent arty s7 50).
The delay is based on artix7 datasheet so maybe to small for the device you use (but this code is used for biggest devices without any problem...).
I have to look at bsd files to see if a longer delay is required or if there is a regression somewhere.

UweBonnes:
Maybe 
 attribute INSTRUCTION_CAPTURE of XC6VLX130T_FF784 : entity is
[...]

PR #440, opened 2024-03-04
xilinx.cpp: After programming, go to bypass, PR for #436

UweBonnes:
Needed for xc7s50 on VMM3 boards to detect FLASH.


trabucayre:
LGTM, just a small question:
An oneshot BYPASS read is enough? Not required to poll until DONE is high?

Thanks!

UweBonnes:
I first has a loop waiting for DONE and a counter for the loop. The loop was only executed once, so I removed the loop and still flash was recognized. Removing the BYPASS instruction resulted again in a fail to recognize the flash.

trabucayre:
Applied Thanks @UweBonnes !

Issue #443, opened 2024-03-13
olimex_gatemateevb: unable to flash


schnommus:
Firstly, thanks for your fantastic work on openFPGALoader thus far!

I have been trying to flash some bitstreams to my Olimex Gatemate A1 board as follows (note: using your published v0.12 release binaries for Ubuntu 22.04):

```
[seb@zenner bin]$ sudo ./openFPGALoader -b olimex_gatemateevb ~/dev/litex-venv/litex/build/olimex_gatemate_a1_evb/gateware/olimex_gatemate_a1_evb_00.cfg.bit 
empty
Jtag frequency : requested 6000000Hz -> real 6000000Hz
Load SRAM via JTAG: [==================================================] 100.00%
Done
Wait for CFG_DONE
```

It hangs on this line 'Wait for CFG_DONE' forever until I terminate it.
Interestingly, the `CFG_DONE_LED1` does light up, so presumably the configuration actually happened successfully (interestingly the LED chaser doesn't flash and I get no UART traffic as one should for a litex example, but that's another kettle of fish... I built with `PATH=/opt/cc-toolchain-linux/bin/p_r:$PATH PATH=/opt/cc-toolchain-linux/bin/yosys:$PATH python3 -m litex_boards.targets.olimex_gatemate_a1_evb --cpu-type vexriscv --build` if that matters).

The more interesting issue I encountered is a segfault on adding the `-f` option:

```
[seb@zenner bin]$ sudo ./openFPGALoader -b olimex_gatemateevb -f ~/dev/litex-venv/litex/build/olimex_gatemate_a1_evb/gateware/olimex_gatemate_a1_evb_00.cfg.bit 
empty
write to flash
Jtag frequency : requested 6000000Hz -> real 6000000Hz
Segmentation fault
```

Is writing to flash supported on this board? It has an SPI flash on it so presumably it is possible but unclear to me if it is supposed to be supported.

Thanks!

NOTE: I also tried re-building and re-flashing dirtyJtag to the MCU from hash 938eb6d if that's relevant

trabucayre:
Thanks for your feedback!

For the first part it's weird: I have used this without being bloked or something similar, but I have to retest.
For the flash aspect I must admit not having tested by myself: I do!.

For LiteX it's also partially weird: sometime p_r produces a bitstream not fully working (but led chaser is always working it's more issues with the CPU/UART).

[...]

PR #445, opened 2024-03-14
gatemate: fix unintended gpio access with dirtyJtag cables

pu-cc:
Refers to issue #443.

JTAG and SPI routines are implemented primarily for FTDI-based configuration and flashing. This commit avoids using the FTDI-GPIOs for monitoring and controlling the `CFG_DONE`,  `CFG_FAILED`, reset and output-enable signals and allows SRAM configuration via JTAG using `dirtyJtag` cables.


trabucayre:
Applied. Thanks @pu-cc !
I will create a minor release as soon as everything is tested using dirtyJtag.

PR #446, opened 2024-03-14
gatemate: fix jtag-spi-bypass with dirtyJtag

pu-cc:
Refers to an issue mentioned in #443.

With this set of commits, external flash can be configured with `dirtyJtag` via the built-in JTAG-SPI-bypass. There are also a few clean-ups, such as no duplicate calls of flash routines.

Tested with `-b gatemate_evb_{spi,jtag}`, `-b olimex_gatemateevb` and `-c dirtyJtag`.


trabucayre:
Awesome!
Applied Thanks @pu-cc.
This was quick to fix this issue!
Thanks.

PR #447, opened 2024-03-14
spiFlashdb: add MX25R6435F and fix SPIFlash::bp_to_len

pu-cc:
This device was missing in spiFlashdb and is installed on the GateMate evaluation board.

MX25R6435F's status register contains a quad-enable bit.
image

If it is set, `SPIFlash::bp_to_len` incorrectly masks all sectors, even if no bp-bit is set. This fix seems obvious to me, but I don't have many boards to test this.

### DIFF
--- a/src/spiFlash.cpp
+++ b/src/spiFlash.cpp
@@ -895,6 +895,9 @@ std::map<std::string, uint32_t> SPIFlash::bp_to_len(uint8_t bp, uint8_t tb)
 	for (int i = 0; i < 4; i++)
 		if ((bp & _flash_model->bp_offset[i]))
 			tmp |= (1 << i);
+	/* 0 -> no block protected */
+	if (tmp == 0)
+		return protect_area;
 	/* bp code is 2^(bp-1) blocks */
 	uint16_t nr_sectors = (1 << (tmp-1));
 	printf("nr_sectors : %d\n", nr_sectors);
--- a/src/spiFlashdb.hpp
+++ b/src/spiFlashdb.hpp
@@ -354,6 +354,20 @@ static std::map <uint32_t, flash_t> flash_list = {
 		.bp_len = 5,
 		.bp_offset = {(1 << 2), (1 << 3), (1 << 4), (1 << 5)}}
 	},
+	{0xc22817, {
+	/* https://www.macronix.com/Lists/Datasheet/Attachments/8868/MX25R6435F,%20Wide%20Range,%2064Mb,%20v1.6.pdf */
+		.manufacturer = "Macronix",
+		.model = "MX25R6435F",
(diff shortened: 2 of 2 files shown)


src/spiFlashdb.hpp
```diff
@@ -354,6 +354,20 @@ static std::map <uint32_t, flash_t> flash_list = {
...
+		.model = "MX25R6435F",
+		.nr_sector = 128,
+		.sector_erase = true,
+		.subsector_erase = true,
+		.has_extended = false,
+		.tb_otp = true,
+		.tb_offset = (1 << 3),
+		.tb_register = CONFR,
+		.bp_len = 4,
+		.bp_offset = {(1 << 2), (1 << 3), (1 << 4), 0}}
```
trabucayre, line 369: `bp_len` is 4 but you only provides 3 positions. Maybe the issue is related to that. I will investigate too.

src/spiFlash.cpp
```diff
@@ -895,6 +895,9 @@ std::map<std::string, uint32_t> SPIFlash::bp_to_len(uint8_t bp, uint8_t tb)
 	for (int i = 0; i < 4; i++)
 		if ((bp & _flash_model->bp_offset[i]))
 			tmp |= (1 << i);
+	/* 0 -> no block protected */
+	if (tmp == 0)
+		return protect_area;
```
trabucayre, line 900: it's true! Its wrong to substract something to 0!

trabucayre:
Applied. Thanks @pu-cc.

I have to check too but maybe the issue is related to the quad mode with a bad read and an flash's answer containing 0xff ?

pu-cc:
Thank you, @trabucayre!

I does not seem that the status register is misread. `0x40` is QE bit: 

```
write to flash
Jtag frequency : requested 10.00MHz  -> real 10.00MHz 
found 1 devices
index 0:
        idcode 0x20000001
        manufacturer colognechip
        family GateMate Series
        model  GM1Ax
        irlength 6
File type : cfg
c2 28 17 c2 read c22817c2
Detected: Macronix MX25R6435F 128 sectors size: 64Mb
RDSR : 40  <- here
WIP  : 0
WEL  : 0
BP   : 0
TB   : 0
SRWD : 0
00000000 00000000 00000000 40
```

We just did not return from `SPIFlash::bp_to_len` if bp != 0, and later subtract to 0:
https://github.com/trabucayre/openFPGALoader/blob/1d276ebb9db52ccd1081fe5bc68bbdc47eeced23/src/spiFlash.cpp#L890

The fix mentioned in https://github.com/trabucayre/openFPGALoader/pull/447#discussion_r1525791830 solves the issues and with the current status everything is perfect for me. Thanks again!

PR #449, opened 2024-03-27
Add support for the Cyclone II

uint69-t:
Add support for the Cyclone II board. **TESTED with RAM**
**NOTES**:
 * Known to work with SVF files

**TODO**:
 * Debug RBF files
 * Debug FLASH load using secondary JTAG header on the board


trabucayre:
Applied. Thanks @uint69-t !

PR #450, opened 2024-03-28
Added Zetta ZD25WQ16CSIGT

kalata23:
This should fix the "Unknown ID" issue


trabucayre:
Applied. Thanks @kalata23 !

PR #453, opened 2024-04-27
Add EP4CGX150

hansfbaier:
I could not test flash, because I am not sure of how to build a bistream for it.
I tried `./build.py ep4cgx15027`, but it errors.
when I cd into the tmp directory of it and run make manually I get the following error:
```
...
Error (12006): Node instance "altclkctrl_component" instantiates undefined entity "altclkctrl_inst". Ensure that required library paths are specified correctly, define the specified entity, or change the instantiation. If this entity represents Intel FPGA or third-party IP, generate the synthesis files for the IP. File: /opt/intelFPGA_lite/22.1/quartus/libraries/megafunctions/alt_sfl_enhanced.v Line: 200
...
```
When I try to build the bistream for EP4CE15 I get the same error,
so I am not sure if that error is my fault.

Loading a bitstream into memory works.

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -79,6 +79,13 @@ Intel:
     Memory: OK
     Flash: OK
 
+  - Description: Cyclone IV GX
+    Model:
+      - EP4CGX150
+    URL: https://www.intel.com/content/www/us/en/products/details/fpga/cyclone/iv/gx/products.html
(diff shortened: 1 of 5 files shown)


trabucayre:
Applied Thanks @hansfbaier.
I observe the same error, I have to verify if something is required or if it's a problem with quartus (version, type, ...).

PR #458, opened 2024-05-19
adding support for XC2C64A-xVQ44 with ID 0x06e5e093

schenkmi:


trabucayre:
Applied. Thanks @schenkmi

PR #459, opened 2024-05-23
xilinx: Add XADC reads of VCC registers

colinoflynn:
This adds support to read the min/max temp, as well as the current+min/max VCCINT and VCCAUX registers (VCCBRAM could be added, but is normally the same as VCCINT).

This can be helpful for troubleshooting devices. I did change the JSON output format slightly to make it more "human friendly" by adding line breaks. However someone parsing the output *should* still be able to use their previous code, provided it can deal with reading in line ends.

```
{"temp": 39.9061, 
    "maxtemp": 40.3194, 
    "mintemp": 25.9852, 
"raw":  {"0": 40684, "1": 21949, "2": 39270, "3": 0, "4": 0, "5": 0, "6": 21948, "7": 0},
"vccint": 1.00415, 
   "maxvccint": 1.00635, 
   "minvccint": 1.00195, 
"vccaux": 1.79663, 
   "maxvccaux": 1.79883, 
   "minvccaux": 1.79517, 
}
```

The current/min/max could be made a dict for each top-level entry too, but I thought that would break existing code (e.g., since "temp" would no longer be a single value). Hopefully this is useful without breaking any use-cases currently!

### DIFF
--- a/src/xilinx.cpp
+++ b/src/xilinx.cpp
@@ -93,6 +93,8 @@
 #define XADC_VAUX13   0x1d
 #define XADC_VAUX14   0x1e
 #define XADC_VAUX15   0x1f
+#define XADC_TEMP_MAX 0x20
+#define XADC_TEMP_MIN 0x24
 #define XADC_SUPBOFFS 0x30
 #define XADC_ADCBOFFS 0x31
 #define XADC_ADCBGAIN 0x32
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied. Thanks @colinoflynn for this improvement!

PR #479, opened 2024-08-12
part.hpp: add Efinix Titanium Ti180

jwise:
Add IDCODE for Efinix Titanium Ti180.  SRAM loading seems to work fine, no bitstream available for SoJ yet.


trabucayre:
Applied. Thanks @jwise 
I'm working to cleanup/integrate build system to produces spiOverJtag bitstream for efinix.

PR #481, opened 2024-08-20
efinix: add spiOverJtag support for Ti180J484

jwise:
while we are in here, beef up the existing efinix.cpp to be able to detect_flash and dump_flash via SoJ, too


src/efinix.cpp
```diff
@@ -228,11 +228,23 @@ void Efinix::program(unsigned int offset, bool unprotect_flash)
...
+bool Efinix::detect_flash()
+{
+	if (_jtag) {
+		return SPIInterface::detect_flash();
+	}
+
+	uint32_t timeout = 1000;
+	_spi->gpio_clear(_rst_pin);
+
+	return reinterpret_cast<SPIInterface *>(_spi)->detect_flash();
```
trabucayre, line 240: Compared to `dumpFlash` here the FPGA isn't reseted after, no?

trabucayre: Have you tested this method in SPI mode?

jwise: I confess I have not tried this in anything other than SoJ mode -- my Ti180 board only supports JTAG. Indeed, it does not reset the FPGA afterwards, I think (unless `post_flash_access` does that?).  Looks like I copied this from `dumpFlash`.  Maybe both `dumpFlash` and this should be refactored to call `reset()`, which does the same thing?  Though as mentioned, I cannot test that on my board, so either way, I would be implementing it blind.

jwise: I've updated this to disallow that path entirely until someone gets to test it.

trabucayre:
Applied. Thanks @jwise !

PR #484, opened 2024-08-25
adding support for cyc5000

gsteiert:
Adding support for CYC5000.
Also updated some Intel documentation for generating rbf and rpd files.


trabucayre:
Applied. Thanks @gsteiert !
Your documentation's update will be usefull for others projects too.

PR #488, opened 2024-09-26
feat: adding support for the xc7z030

shareefj:


trabucayre:
Applied. Thanks @shareefj !

PR #489, opened 2024-10-01
Add xc2c64a QFN48

mfacton:
Added ID for XC2C64A QFN48 chip.


trabucayre:
Applied. Thanks @mfacton

PR #490, opened 2024-10-10
Add ID and spiOverJtag bitstream for Stratix V GS D5

G33KatWork:
Hi,
I added support for a Stratix V GS D5 FPGA.

Specifically this PR is for the support of the Microsoft Azure Catapult V2 cards named "Pikes Peak" and "Storey Peak". Those cards, which show up cheaply on eBay every now and then, contain a special version of a Stratix V GS D5 with the full model number of `5SGSMD5K1F40C1` and a JTAG device ID of `0x029070dd`.

The `spiOverJtag` design that I built is for `5SGSMD5K2F40I3` which should be the closest match that is not "special" and publicly documented and it works on my FPGA as well. The built bitstream might even work on all Stratix V GS D5 FPGAs, however I have not tested that. Also, to build the bitstream you need at least Quartus Standard. The free Lite version is not enough.

I tested the FPGA SRAM write and SPI flash read as follows on a Storey Peak card and got good results:

```
./openFPGALoader -c ft2232 --vid 0x0403 --pid 0x6014 --fpga-part 5SGSD5 --dump-flash --file-size 33554432 -B ../spiOverJtag/spiOverJtag_5sgsd5.rbf.gz --bitstream foo.bin
./openFPGALoader -c ft2232 --vid 0x0403 --pid 0x6014 --fpga-part 5SGSD5 gateware/build/storey_peak/gateware/storey_peak.rbf
```

edit:
Oh, one more thing! I had to patch something in the quarus megafunctions files, because I wasn't able to build designs that use the serial flash loader. I always got the following error message: `Node instance "altclkctrl_component" instantiates undefined entity "altclkctrl_inst".`

When googling, I found the following Intel Community forum thread: https://community.intel.com/t5/Programmable-Devices/Error-Message-12006-with-Quartus-Prime-Standard-Edition-23-1/m-p/1572452

The official help wasn't particularly helpful, so I just winged it and patched line 195 in file `quartus/libraries/megafunctions/alt_sfl_enhanced.v` where this `altclkctrl` block is instantiated from `altclkctrl_inst altclkctrl_component` to `altclkctrl altclkctrl_inst` and stuff started to work. Just FYI, this seems to be an issue on Intel/Altera's side.

All this on Quartus Prime Standard 23.1.


trabucayre:
Applied. Thanks @G33KatWork 
And thanks for informations about work around required with megafunction!

PR #493, opened 2024-10-17
Update Max II and Cyclone II documentation

Tpj-root:
Add documentation for Max II and Cyclone II boards

- Included details on tested device models: EP2C5T144C8N and EPM240T100C5N


trabucayre:
Applied. Thanks @Tpj-root !

PR #496, opened 2024-11-12
Add support for VCU108 board and Virtex UltraScale

schodet:
Tested by loading a design on the VCU108 board.

No flash support for the moment.


trabucayre:
Applied. Thanks @schodet !

PR #497, opened 2024-11-26
src/part.hpp: Add Altera Max10 10M16SAU169C8G support (Used by LimeSD…

enjoy-digital:
…R Mini V1).


trabucayre:
Applied. Thanks @enjoy-digital !

PR #498, opened 2024-12-02
doc/guide/install.rst: instructions to install under Guix

csantosb:
Open fpga loader is [available](https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/fpga.scm?h=master#n609), operational and up to date in the guix standard repositories. The advantage about other package managers is that users don't need to bother about dependencies, as they get installed along with it.


trabucayre:
Applied. Thanks @csantosb

PR #499, opened 2024-12-10
gowin: Implement User Flash programming for GW1N9

jeanthom:
Hi,

Gowin FPGA with built-in flash memory offer a separate "user flash" zone ([documentation](https://cdn.gowinsemi.com.cn/UG295E.pdf)), either accessible from programmable logic, or used as system memory for the EMCU. This memory can be configured by JTAG, but this isn't documented at the moment. We've reversed how flashing was performed for the GW1N9 FPGAs by sniffing the JTAG wires: it works in the same way you flash the bitstream, but with a different base address.

This PR implements user flash programming for GW1N(R)9 FPGAs (other models probably have different base addresses). The documentation lists this addition as experimental because of the reverse engineered nature of this development. A few safeguards have been implemented to prevent misuse of this functionality (FPGA model checking, file size).

Jean

### DIFF
--- a/src/gowin.cpp
+++ b/src/gowin.cpp
@@ -77,8 +77,8 @@ using namespace std;
 
 Gowin::Gowin(Jtag *jtag, const string filename, const string &file_type, std::string mcufw,
 		Device::prog_type_t prg_type, bool external_flash,
-		bool verify, int8_t verbose): Device(jtag, filename, file_type,
-		verify, verbose),
+		bool verify, int8_t verbose, const std::string& user_flash)
+	: Device(jtag, filename, file_type, verify, verbose),
 		SPIInterface(filename, verbose, 0, verify, false, false),
 		_idcode(0), is_gw1n1(false), is_gw1n4(false), is_gw1n9(false),
 		is_gw2a(false), is_gw5a(false),
--- a/src/gowin.cpp
+++ b/src/gowin.cpp
@@ -80,12 +80,11 @@ Gowin::Gowin(Jtag *jtag, const string filename, const string &file_type, std::st
 		bool verify, int8_t verbose): Device(jtag, filename, file_type,
 		verify, verbose),
 		SPIInterface(filename, verbose, 0, verify, false, false),
-		_fs(NULL), _idcode(0), is_gw1n1(false), is_gw2a(false),
-		is_gw1n4(false), is_gw5a(false), _external_flash(external_flash),
+		_idcode(0), is_gw1n1(false), is_gw2a(false), is_gw1n4(false),
+		is_gw5a(false), _external_flash(external_flash),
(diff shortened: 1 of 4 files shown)


jeanthom:
Looking at the diff from the PR, it seems like my attempt at fixing indentation issues modified more core than I expected. Should I remove those lines from this PR?

src/gowin.cpp
```diff
@@ -97,6 +98,30 @@ Gowin::Gowin(Jtag *jtag, const string filename, const string &file_type, std::st
...
+
+		if (_uf->parse() == EXIT_FAILURE) {
+			printError("FAIL");
+			delete _mcufw;
+			throw std::runtime_error("can't parse user flash file");
+		}
+
+		const int kPageSize = 64 * 32;
+		int num_pages = (_uf->getLength() + kPageSize - 1) / kPageSize;
+		if (is_gw1n9 && num_pages > 304)
```
trabucayre, line 121: no `delete` here ?

trabucayre:
It looks good, but I have a question: is mcufw principle is not working in this case? The approach here looks similar to the one for 4k.
I'm wrong?

Thanks

src/gowin.cpp
```diff
@@ -97,6 +98,30 @@ Gowin::Gowin(Jtag *jtag, const string filename, const string &file_type, std::st
...
+
+		try {
+			_uf = new RawParser(user_flash, false);
+		} catch (std::exception &e) {
+			throw std::runtime_error(e.what());
+		}
+
+		if (_uf->parse() == EXIT_FAILURE) {
+			printError("FAIL");
+			delete _mcufw;
```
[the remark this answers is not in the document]
jeanthom: Renamed into `_userflash`

src/gowin.cpp
jeanthom: Memory management replaced by unique_ptr

jeanthom:
Hi,

I just force pushed a newer version of this MR. I kept separate arguments for mcufw and userflash to avoid confusion (Gowin documentation clearly distinguishes between the two). Let me know what you think?

trabucayre:
Globally this PR is ready to be merged.
In fact the only think where I'm hesitant is with the addition to a new argument: openFPGALoader has many options and I have no idea about the part of time took by  the argument parser compared to the operation. Also for user it's maybe hard to find which argument to use.

jeanthom:
I don't really think the argument parsing is much of a concern here. I can run some benchmarks if you want though.

I understand your concerns regarding the user experience being affected by extra arguments. However I can't think of a better solution, that's both doesn't require complicated setup and that would be easy to discover.

trabucayre:
Applied Thanks @jeanthom!
I will do some tests regarding arguments and maybe trying to reorder it to have something more easy

PR #500, opened 2024-12-21
Debian/Ubuntu instructions for binary installation. Update install.rst

vlad1k337:
openFPGALoader binary is available in Debian/Ubuntu repositories. Added installation instructions for these distributions. New Linux users may be confused by necessity to compile the program, added installation instructions for binary packages.


trabucayre:
Applied. Thanks @vlad1k337

PR #502, opened 2024-12-31
fix: update version to patch with the release

chenrui333:
👋 seeing the [version mismatch doing the release build for 0.13.0](https://github.com/Homebrew/homebrew-core/pull/202864), bumping the version to match with the release. Thanks!

cc @trabucayre


trabucayre:
Good catch! Once again I have missed to update version...
Thanks @chenrui333

chenrui333:
np, it happens. <3

PR #505, opened 2025-01-08
Add Altera 10M02SCM153C8G Support.

steward-fu:
Description:
STEP-MAX10 development board is equipped with Altera 10M02SCM153C8G. It is not in the support list. This PR adds this chip support. I have tested this change with my STEP-MAX10 development board. It runs as expected result. If this PR needs more inforamtion, please let me know, thanks.

Change:
Added 'Altera 10M02SCM153C8G' in fpga_list.


trabucayre:
Applied. Thanks @steward-fu !
Feel free to open a PR for the board too :)

PR #509, opened 2025-01-12
doc/FPGAs.yml: Fix yml structure in Intel Max 10

Frzoen:
```Model: 10M02, 10M08``` 
is not correct yaml structure and parsing it generated `10m02,_10m08` (single) model instead of `10m02` and `10m08` models


trabucayre:
Good catch! Applied. Thanks @Frzoen !

PR #512, opened 2025-01-28
Fix snprintf usage

mtdudek:
Some code used snprintf on std::string objects.
This caused unexpected behaviour where '\0' was printed to the output file. 
This pull request adds intermediate char* buffer which is later used to initialize std::string.
This string is later resized to the correct expected length.


trabucayre:
Applied. Thanks @mtdudek !

PR #527, opened 2025-03-07
gowin: Fix compilation on musl

mc-requtech:
`gowin.cpp` uses the functions `htole32()` and `le32toh()` without including the prerequisite header `<endian.h>`.  This breaks the build when using the musl libc.


trabucayre:
Applied. Tahnks @mc-requtech !

PR #528, opened 2025-03-10
libgpiodJtagBitbang: add support for XVC server

mc-requtech:
Adds the method `writeTMSTDI()` to the libgpiod bitbang driver, making it usable for the XVC server mode.

I have only tested the APIV2 code, but the other one should work for symmetry reasons.  :smile_cat:


trabucayre:
Applied. Thanks @mc-requtech

PR #530, opened 2025-03-21
Added support for MX25L25645G FLASH memory

vbuitvydas:
Adding support for MX25L25645G SPI FLASH to set quad mode bit. Tested on hardware.


trabucayre:
Applied. Thanks @vbuitvydas !

PR #535, opened 2025-03-30
make ENABLE_OPTIM non effective for Debug builds configuration

phdussud:
Just a suggestion. I was surprised when my debug build was not debug...


trabucayre:
Make sense!
Applied. Thanks @phdussud

PR #537, opened 2025-04-09
Add support for xc7k480t

pigmoral:
Hi, this added support for loading bitstream into xc7k480t SRAM.

### DIFF
--- a/doc/FPGAs.yml
+++ b/doc/FPGAs.yml
@@ -260,6 +260,7 @@ Xilinx:
       - xc7k325t
       - xc7k410t
       - xc7k420t
+      - xc7k480t
     URL: https://www.xilinx.com/products/silicon-devices/fpga/kintex-7.html#productTable
     Memory: OK
     Flash: OK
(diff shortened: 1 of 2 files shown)


pigmoral:
Besides, I tried to add spiOverJtag support (https://github.com/pigmoral/openFPGALoader/tree/xc7k480t-support-flash), but failed. 

The flash model on my board is MT28GU512AAX1E-BPI-X16 (Data Width Bits only supports x16 according to UG908). The problem is I get different `Jedec ID`s every time I use the -f option, and I have no idea how to fix it for now.

trabucayre:
Applied. Thanks @pigmoral 
For the Flash: have you a board schematic to share?
Thanks!

pigmoral:
Hi @trabucayre ,

> For the Flash: have you a board schematic to share?

The board I use is an accelerator card that leaked into the used market in China. There is no schematic available, but some constraints obtained through hacking can be found at:
https://github.com/TiferKing/ypcb_00338_1p1_hack/blob/main/constraints/ypcb003381p1.xdc

PR #538, opened 2025-04-11
Add support for Artix A7 15t

nate-nexthop:
Tested on real hardware, connecting through Linux gpiod.

SRAM and QSPI loading were successful.


trabucayre:
Applied. Thanks @nate-nexthop

PR #545, opened 2025-04-27
Add support for Digilent Anvyl, the Big Red Board.

duskwuff:
Programming to SRAM works. Programming flash works with a renamed spiOverJtag bitfile - cs(g)484 is different from fg(g)484, but the configuration flash IOs all end up on the same pads in the end.

(As an aside: I suspect that more Xilinx bitstreams are being generated than are actually needed. There's no physical difference between the silicon for e.g. xc6slx45csg324, xc6slx45fgg484, and xc6slx45csg484 - the configuration flash is connected to the same pads on each part, the pins are just named differently. As such, the same bitstream can be used for any package containing that part.)


trabucayre:
Applied. Thanks @duskwuff 
Please updates doc too to list your board ;-)
Indeed, it's seems possible to use the same bitstream for the full model/size. I'm working on a spiOverJtag v2 and have to integrates this to reduces amount of bitstreams.
Unfortunately I haven't enough boards with the same model but with a different size to be sure for all FPGAs requiring spiOverjtag.

PR #547, opened 2025-05-08
Add xcau15p support

Tomserv-512:
These commits add support for Ultrascale+ device xcau15p.
An error in xilinx_spiOverJtag.v when _xilinxultrascale_ is defined, but _secondaryflash_ is not is also fixed.


trabucayre:
Applied. Thanks @Tomserv-512 !

PR #563, opened 2025-06-10
Add cable support for ARM-USB-TINY-H based off ARM-USB-OCD-H

### LINKED ISSUE #562: Add support for Olimex ARM-USB-TINY-H
Remarkably similar to the Olimex ARM-USB-OCD-H: they have the same pinout and I can confirm `openFPGALoader -c arm-usb-ocd-h --vid 0x15ba --pid 0x002a` already works with it. It should be a matter of adding an entry for it into `src/cable.hpp` by copy-pasting the existing entry for `arm-usb-ocd-h`.

jmi2k:
Fixes #562. The existing code suffices to use the new device as a JTAG programmer.

This assumption has been tested by forcing the VID:PID pair using `--vid 0x15ba --pid 0x002a`, but the PR itself remains untested until I have access to the device (this Thursday).


trabucayre:
Applied. Thanks @jmi2k

omixann:
Thanks. Just for the record: it works and is able to detect and program my FPGA.

image

PR #568, opened 2025-07-03
Document use of udev rules under GuixSystem

csantosb:
Udev rules are [included](https://codeberg.org/guix/guix/commit/690ac4fc2d) in openfpgaloader guix package.

This pr aims at documenting its use.


trabucayre:
Applied. Thanks @csantosb !

PR #571, opened 2025-07-10
colognechip: Simplify handling of JTAG bypass bits

pu-cc:
This PR simplifies the handling of bypass bits with multiple GateMate FPGAs in JTAG chains.

Tested with A1, A2 and various combinations of JTAG chains as well as with the official and the Olimex evaluation kits.


trabucayre:
Applied. Thanks @pu-cc !

PR #580, opened 2025-09-03
spiFlash: mask RDSR_WIP instead of RDSR_WEL while waiting for completion

pu-cc:
Just a very minor fix:

During testing with [AT25QL321](https://www.mouser.at/datasheet/3/1166/1/REN_DS-AT25QL321-131E-022023_DST_20230327_2.pdf) for ULX5M, I noticed that when changing the quad enable bit, we wait for the write enable latch (WEL) instead of the busy flag (WIP).

This resulted in the loop already being interrupted, even though the busy flag was still active.


trabucayre:
Applied. Thanks @pu-cc !

PR #581, opened 2025-09-03
spiFlashdb: add AT25QL321

pu-cc:
This flash is used in https://github.com/intergalaktik/ulx5m-gs.


trabucayre:
Applied. Thanks @pu-cc

PR #582, opened 2025-09-03
colognechip: add bulk erase + quad enable features and simplify spi_wait overload

pu-cc:
The routines for bulk erase and quad bit modification were still missing.

Moreover, `CologneChip::spi_wait` had an unnecessary condition. 
ULX5M flashing with AT25QL321 works after these changes. Tested with official evalboard, olimex board and ULX5M.


trabucayre:
Perfect.
Applied. Thanks @pu-cc !

PR #584, opened 2025-09-07
Add altera EP4CE30 

drandyhaas:


trabucayre:
Applied. Thanks @drandyhaas !

PR #591, opened 2025-10-02
Update part.hpp to support Lattice LFE3-150EA device

hennomann:
This PR adds device ID support for the Lattice ECP3 LFE3-150EA. Previously, only the LFE3-70E was supported.

We've tested the implementation and confirmed it works as expected. We're happy to contribute to expanding ECP3 device support!


trabucayre:
Applied. Thanks @hennomann !

Issue #560, opened 2025-06-04
Does openFPGALoader support the Xilinx Versal family


orodkey:
Hi 

I would like to know if this FPGAloader supports the Xilinx Versal devices as well ? or if you can point me to the part of the code to add it there.
I'm trying to write a .pdi file into the flash !

I tried these 

~/fpga# ./openFPGALoader -c ft4232 -f /root/Orod/boot.pdi

Jtag frequency : requested 6.00MHz   -> real 6.00MHz
Open file DONE
Parse file DONE
Can't program SPI flash: missing device-package information

~/fpga# ./openFPGALoader -c ft4232 --index-chain 1 --freq 7500000 /root/Orod/boot.pdi
Jtag frequency : requested 7.50MHz   -> real 7.50MHz
Error: device 5643 not supported

This is the Flash chip that we're using MT25QL256ABA1.

Thanks,
Orod

trabucayre:
Versal are not, officially supported by openFPGALoader (maybe someone has code for that).
But since [the same code base](https://github.com/trabucayre/openFPGALoader/blob/master/src/xilinx.cpp) is working for, more or less, all Xilinx its looks reasonable to think it will be okay for this family. Xilinx provides documentations on how to load a bitstream in the SRAM.

For the flash access it's another question because [a bridge](https://github.com/trabucayre/openFPGALoader/tree/master/spiOverJtag) is required between JTAG interface and SPI Flash interface: bitstreams are available for Spartan X, artix and kintex but not for the Versal family.

Since  `pdi` file is not supported, [a parser](https://github.com/trabucayre/openFPGALoader/blob/master/src/configBitstreamParser.cpp) must be written to adds support for this format.

AEW2015:
I plan on look at this soon.  Just create a basic sequence to send the PID to the correct JTAG IR code.  This will be to support Versal and Spartan Ultrascale+.

AEW2015:
So I have been playing around with it and I can't even get the xvc feature of openFPGALoader to program the spartan ultrascale plus through vivado...

It seems the PDI and PMC is a dramatic difference that will take more work.  I try to have an udpate next week.

AEW2015:
Okay I got a tcl script to do a basic prog_pdi for Spartan Ultrascale Plus... I will try porting that over to openFPGALoader for basic pdi programing.  I can look at parsing the PDI more later.
[...]

PR #593, opened 2025-10-14
basic PDI prog for Spartan Ultrascale+

AEW2015:
Related to https://github.com/trabucayre/openFPGALoader/issues/560

This adds basic support for PDI for the Spartant Ultrascale+.
This should work on other PDI systems (like versal), but needs testing.

Might be good to break out the new JTAG STATUS register better, then me hard-coding it....


trabucayre:
Thanks for this PR.
Just few questions:
- `program_pdi` looks really similar to `program_mem`: what do you think to merge both in `program_mem` with the use of device type to bypass/execute some portions of code (I'm agree this adds a bit of extra logic but it's more easy to maintain in a long term)
- the `ircode_mapping` follows default registers code: maybe it's also possible to use it instead of adding a new sequence?
- `PDI_MODE` is maybe not required because this one mean `MEM_MODE && _fpga_family == SPARTANUSP_FAMILY`

AEW2015:
Yea I can do some cleanup and try to reduce as much as possible.  Will test on my HW.

src/xilinx.cpp
```diff
@@ -358,6 +361,14 @@ Xilinx::Xilinx(Jtag *jtag, const std::string &filename,
 		_fpga_family = KINTEXUSP_FAMILY;
 	} else if (family == "artixusp") {
 		_fpga_family = ARTIXUSP_FAMILY;
+	} else if (family == "spartanusp") {		
+		if (_file_extension != "pdi") {
+			char mess[256];
+			snprintf(mess, 256, "Error: only voltaile PDI programing for "
```
trabucayre: typo `voltaile` -> `volatile`

trabucayre:
Applied. Thanks @AEW2015 !

PR #594, opened 2025-11-07
Bit-file validation 

debuti:
The application fails "successfully" (exit code 0) when a non existing bit_file is passed as parameter


trabucayre:
I see one issue with this approach: the bit_file may be a existing file to load/write or a (potentially) no-existant file when a dump must be performed.

trabucayre:
Any news?
Another issue with this code: it's possible to uses stdin with '|' to provides bitstream.

PR #597, opened 2025-12-11
spiOverJtag: Add support for Efinix T120F324

zh522130:
Tested on T120F324C4 chip and works correctly.


trabucayre:
Applied.
Thanks @zh522130 !

PR #598, opened 2025-12-12
efinix: Add header parsing and flash programming validation

zh522130:
Parse bitstream header to extract Mode, Width and Device fields. Add validation for flash programming:
- Check device matches --fpga-part parameter
- Reject passive mode (only active mode supported for flash)

This prevents flashing incorrect bitstreams that would fail to boot.

I tested with the T120F324 and also extracted SPI-over-JTAG bit files for T8F81, T13F256, TI60F255, and TI180J484 for verification.

### DIFF
--- a/src/efinix.cpp
+++ b/src/efinix.cpp
@@ -8,6 +8,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <algorithm>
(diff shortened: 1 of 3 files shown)


trabucayre:
Applied.
Thanks @zh522130

PR #599, opened 2025-12-12
anlogic: fix high frequency flash programming issue

zh522130:
When programming the flash on my `EG4S20BG256` at `15 MHz`, an error occurs. Details:  
```
Jedec ID: ff  
Memory type: ff  
Memory capacity: ff  
Flash chip unknown: use basic protection detection  
Unlock blocks  
Error: block protection is set.  
Cannot unlock without --unprotect-flash  
```
This issue arises because reading the ID cannot be performed at too high a clock frequency. I resolved this by limiting the clock frequency for reading the ID. The official programming tool also adjusts the frequency dynamically, though it is more conservative than my code—I recall it drops to 3 MHz. The 6 MHz frequency was set as a reference to the original default frequency. This can be adjusted later if any issues arise.


trabucayre:
Applied.
Thanks @zh522130 !

PR #604, opened 2025-12-24
Enable support for CH347F

phdussud:
The CH347F variant has a different PID because it isn't code compatible with the T variant. The bulk interface to target JTAG is different. Rest is the same. 
Fix a Windows issue: libusb_set_auto_detach_kernel_driver is meaningless in Windows and it never return a success code.


trabucayre:
Applied. Thanks @phdussud

phdussud:
Merry Christmas!

PR #606, opened 2025-12-27
dirtyjtag: Allow custom VID/PID via command line options

germaneguise:
## Summary
  Pass cable.vid and cable.pid to DirtyJtag constructor instead of hardcoded values.

  This allows users to use DirtyJTAG-compatible firmware with custom USB VID/PID:
  openFPGALoader -c dirtyJtag --vid 0x1337 --pid 0x0001 bitstream.fs

  ## Use case
  - Custom DirtyJTAG implementations on embedded MCUs
  - Testing with different VID/PID combinations

  ## Changes
  - `dirtyJtag.hpp`: Add VID/PID parameters with defaults (0x1209:0xC0CA)
  - `dirtyJtag.cpp`: Use passed VID/PID instead of hardcoded defines
  - `jtag.cpp`: Pass cable.vid/pid to constructor
  - `xvc_server.cpp`: Same change for XVC server

  Backward compatible - default VID/PID unchanged.

### DIFF
--- a/src/dirtyJtag.cpp
+++ b/src/dirtyJtag.cpp
@@ -62,7 +62,7 @@ enum dirtyJtagSig {
 	SIG_SRST = (1 << 6)
 };
 
-DirtyJtag::DirtyJtag(uint32_t clkHZ, int8_t verbose):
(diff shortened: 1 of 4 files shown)


trabucayre:
Applied.
Thanks @germaneguise !

PR #608, opened 2026-01-07
Fix a small formatting error

lipx1508:
`obuf - _obuf` returns `long` rather than `long long`


trabucayre:
Applied.
Thanks @lipx1508 !

PR #612, opened 2026-01-18
spiFlashdb: Added MX25V8035F NOR flash chip

Wolfiwolf:


trabucayre:
Applied.
Thanks @Wolfiwolf !

PR #617, opened 2026-01-22
Add xc7z007s

fisherdog4:
Checked on my own device and appears to be corroborated here: https://libera.catirclogs.org/openFPGALoader/2024-06-08#


trabucayre:
Applied.
Thanks @fisherdog4 
Feel free to also adds board definition

PR #619, opened 2026-01-28
xilinx: Fixed infinite loop in bit files smaller than 100

rkebelj:


trabucayre:
Applied.
Thanks @rkebelj !

PR #621, opened 2026-02-03
Add spiOverJtag for 10CL006YE144C8G + fix for quartus 25

Ced2911:
Add spiOverJtag for 10CL006YE144C8G
+ Fix based on information from: https://github.com/trabucayre/openFPGALoader/pull/453


trabucayre:
Good catch!
Applied.
Thanks @Ced2911 !

PR #623, opened 2026-02-08
Add workaround for slow JTAG on Sipeed Console

gatecat:
Not sure if I implemented this in the neatest way possible, but at least the logic is here ^^

The MCU on the Sipeed Console apparently has a bug where LSB first mode is incredibly slow (eventually found by comparing wireshark traces between gowin programmer and openfpgaloader). Before this patch it was taking about 4 minutes to program the SRAM of the 138k, with this patch it takes 14 seconds...

### DIFF
--- a/src/ftdiJtagMPSSE.cpp
+++ b/src/ftdiJtagMPSSE.cpp
@@ -80,6 +80,13 @@ void FtdiJtagMPSSE::init_internal(const mpsse_bit_config &cable)
 		_cmd8EWA = true;
 	}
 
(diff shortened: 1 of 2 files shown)


trabucayre:
I usually not really happy when it's required to add work around but I think there is no choice here.
It maybe relevant to raise a message to sipeed about this limitation / error.
Just a small nitpick before applying.
Thanks

trabucayre:
Applied.
Thanks @gatecat !

PR #624, opened 2026-02-10
Adding proper support for YPCB-00338-1P1 and DU-PCB-001-003 (YZCA variant)

blurbdust:
This adds BPI flash programming support to openFPGALoader for the YPCB-00338-1P1 Kintex-7 accelerator card (MT28GU512AAA, 512Mbit)

A new JTAG bridge core (`bpiOverJtag_core.v`) drives the BPI control signals from JTAG USER1 shifts, same approach as the existing spiOverJtag bridges. The bridge implements single-word read, single-word write, and burst write commands.
`BPIFlash` handles Intel CFI flash operations: device detection, block erase, buffered programming (0xE9 command sequence), status polling, and read-back verification. The data path applies the same bit-reversal and byte-swap transformations that Vivado's `write_cfgmem` uses for BPIx16 configuration.
Board integration hooks into the existing Xilinx flow. When `-b ypcb003381p1` is specified with `-f`, it loads the BPI bridge bitstream instead of an SPI bridge, instantiates `BPIFlash`, and routes through `program_bpi()` instead of `program_spi()`.

The first attempt of one USB round-trip per 16-bit word takes about 26 minutes for an 18.7MB bitstream so tthere is a burst write mode that accepts a word count followed by N data words in a single DR shift, with the address auto-incrementing between words. The 21 padding bits after each data word provide the clock cycles the flash needs for its write timing (20 execution cycles plus one for the state transition back). This brings the whole 512-word buffered program operation down to a single USB round-trip.

New files:
- `bpiOverJtag/bpiOverJtag_core.v` -- JTAG-to-BPI bridge FSM
- `bpiOverJtag/xilinx_bpiOverJtag.v` -- top-level wrapper instantiating BSCANE2 primitives
- `bpiOverJtag/constr_xc7k480t_bpi_ffg1156.xdc` -- pin constraints for the xc7k480t
- `bpiOverJtag/build_bpi_xc7k480t.tcl` -- Vivado tcl script
- `bpiOverJtag/bpiOverJtag_xc7k480tffg1156.bit.gz` -- pre-built bridge bitstream
- `src/bpiFlash.hpp` / `src/bpiFlash.cpp` -- BPI flash driver
- Board entry in `doc/boards.yml`, `src/board.hpp`
- Integration in `src/xilinx.cpp` / `src/xilinx.hpp`


trabucayre:
Great works!
I have to perform a closer look and already I see some simplifications / improvements. It is possible to avoid to have this mode overriden for this specific device: the device may also be used in SPI and others devices may be used in BPI.

trabucayre:
Hi.
I have created some commit around your PR [here](https://github.com/trabucayre/openFPGALoader/tree/bpiOverJtag)
What do you think about these modifications?

trabucayre:
Applied.
Thanks @blurbdust !

PR #627, opened 2026-02-17
Fix cross compile on ubuntu 24.04.4 and add clang cross compile support

selimsandal:


trabucayre:
Applied.
Thanks @selimsandal

PR #630, opened 2026-02-24
Small fixes to the compatibility lists

cyrozap:
As I was reading through the docs for the first time, I saw a few small issues in the compatibility lists that could be fixed.

The first was that the abbreviated terms were not sorted alphabetically. Even though they're short lists, sorting them in alphabetical order makes it easier for the reader to quickly find the term they're looking to know the definition of.

The second issue was that the abbreviation "EF" was not expanded in the FPGA compatibility list.

This PR fixes both of those issues.


trabucayre:
Applied.
Thanks @cyrozap !

PR #631, opened 2026-03-01
Allow building XVC client without libftdi

smarsching:
Unlike the XVC server, the XVC client does not depend on libftdi, but due to using a common CMake variable (`ENABLE_XILINX_VIRTUAL_CABLE`), it wasn’t possible to just build the XVC client.

This PR separates this variable into two new variables (`ENABLE_XILINX_VIRTUAL_CABLE_CLIENT` and `ENABLE_XILINX_VIRTUAL_CABLE_SERVER`), so that the XVC client can be built when libftdi is not present.


trabucayre:
Applied.
Thanks @smarsching !

PR #635, opened 2026-03-03
Add SPI over JTAG bitfile for XC7K325TFBG900

smarsching:
This PR adds SPI over JTAG support for the Kintex 7 XC7K325TFBG900.

While this device is largely similar to the XC7K325TFFG900, the bitfile for that device did not work in my tests, while the one specifically built for the XC7K325TFBG900 does.


trabucayre:
Applied.
Thanks @smarsching !
Yes I have observed, unlike Artix where for a die the package is not relevant, for kintex the exact bitstream must be used.

PR #636, opened 2026-03-16
Add XCKU095 support

kevmo314:
I verified this by shipping an LED example over to my XCKU095 and it runs correctly.

Before

```
JTAG init failed with: Unknown device with IDCODE: 0x23844093 (manufacturer: 0x049 (Xilinx), part: 0x1c vers: 0x2  
```

After

```
$ sudo openFPGALoader --cable ft232 --detect
empty
Jtag frequency : requested 6.00MHz    -> real 6.00MHz   
index 0:
        idcode 0x23844093
        manufacturer xilinx
        family kintexus
        model  xcku095
        irlength 6
```


trabucayre:
Applied.
Thanks @kevmo314 !

PR #637, opened 2026-03-23
Add support for Altera MAX10 10M16D IdCode:0x31030dd

EmmanuelP:


trabucayre:
With svf this PR is enought
But to uses native support [this array](https://github.com/trabucayre/openFPGALoader/blob/master/src/altera.cpp#L322-L351) must be updated too.
It's a bit boring to do because informations comes from bsdl file

EmmanuelP:
> With svf this PR is enought But to uses native support [this array](https://github.com/trabucayre/openFPGALoader/blob/master/src/altera.cpp#L322-L351) must be updated too.

Ok, we will provide this array too.

> It's a bit boring to do because informations comes from bsdl file

How are we supposed to extract the information from this file ? Can't they be found in the documentation ?

trabucayre:
In fact nothing is mentionned in documentation, to find this informations I have crossed bsdl file and svf.
I have to update the documentation page to explain how to extracts informations from these files.

EmmanuelP:
I have added memory map information. They are the same as the one for 10M16SA.

We were able to successfully program the flash using this patch and the command:

```
./openFPGALoader -c libgpiod --pins 11:25:10:9 max10_firmware_r1.pof
```

On a MAX10 10M16DCF256.

trabucayre:
Applied.
Thanks @EmmanuelP !

PR #638, opened 2026-03-25
src/colognechip.cpp: allow jtag-spi-bypass for A2

pu-cc:
Since the A2 has two JTAG taps in a chain and the A4 has four, the TDO must be shifted in accordance with the bypass bits when reading data from the JTAG-SPI-bypass / when updating the external flash via JTAG. This update for A2 was a little overdue.

Tested with A1, A2 and multiple programmers/cables such as FTDI and dirtyJtag.


trabucayre:
Applied.
Thanks @pu-cc !

PR #645, opened 2026-04-09
add Adesto AT25DF321A support

ND-00:
Add Adesto AT25DF321A (JEDEC 0x1f4701). Tested on GW2A-LV18QN144 with ft2232. Previously detected as unknown, unprotect failed. Fixed with bp_len=3 and global_lock=false.


trabucayre:
Applied.
Thanks @ND-00 !

PR #647, opened 2026-04-12
vivado 2025.2.1 compatibility: add SW_CRC field to bitstream parser

chop0:
A new field, SW_CRC, was added to BIT file headers in Vivado 2025.2.1;  this patch adds it to bitparser.cpp so it doesn't cause an 'Unknown Key' error.


trabucayre:
Applied.
Thanks @chop0 !

PR #649, opened 2026-04-19
xvc server: increase buffer size to 1MiB for improved network performance

chop0:


trabucayre:
Applied.
Thanks @chop0 !

PR #650, opened 2026-04-20
bitparser: Relax unknown keys from error to warning

zyp:
I have a couple of bitstreams with the following headers, and was getting an error due to the unknown `COMPRESS` key:
- `sqrl_acorn;COMPRESS=TRUE;UserID=0XFFFFFFFF;Version=2024.1`
- `alibaba_xcku3p;COMPRESS=TRUE;UserID=0XFFFFFFFF;Version=2024.1`

The rejection of unknown keys was added in 63c9741, and has already caused one other issue (#647), so I propose relaxing the error to a warning, in addition to adding the `COMPRESS` key.

### DIFF
--- a/src/bitparser.cpp
+++ b/src/bitparser.cpp
@@ -112,8 +112,7 @@ int BitParser::parseHeader()
 						} else if (key == "SW_CRC") {
 							_hdr["crc"] = value;
 						} else {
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied.
Thanks @zyp !
Yes a simple warning make sense.

PR #651, opened 2026-04-23
Enabling the compilation on OpenBSD

dettus:
I took the liberty of trying to compile openFPGAloader on OpenBSD. 
After adding one include, ItWorkedOnMyMachine(tm).

I also added a section to the Readme.md file, allowing interested users to disable a problematic module inside the kernel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,41 @@ OPENFPGALOADER_SOJ_DIR=/somewhere openFPGALoader xxxx
 `OPENFPGALOADER_SOJ_DIR` must point to directory containing **spiOverJtag**
 bitstreams.
 
+## Disabling uftdi on OpenBSD
+
+Certain evaluation boards can cause the following error when running openFPGAloader on OpenBSD:
+
+```
+fail to read data usb bulk read failed
+JTAG init failed with: low level FTDI init failed
+```
+
+This issue is most likely caused by the uftdi module, trying to access the device. To disable it,
+the following commands can be used:
+
+```bash
+# doas config -e -f -o /bsd.nouftdi /bsd
+OpenBSD 7.8 (GENERIC) #54: Sun Oct 12 12:45:58 MDT 2025
+    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/GENERIC
+Enter 'help' for information
+ukc> disable uftdi*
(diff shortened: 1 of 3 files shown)


trabucayre:
The `xvc_server.cpp` looks good.
But for details: it make sense to have this in `doc/guide/install.rst`. Also if you disables ftdi module you will lost ability to have access to uart port. Maybe there is something like runtime unload as done for linux?

dettus:
No, with OpenBSD, it is not a loaded Kernel module. It is hard-linked.
So, unfortunately, there is no easy way to "unload" it, sorry. 
You are indeed correct about the UART, though.
(I am currently waiting for OpenBSD 7.9 to come out next week, then I will have a look into the kernel sources. Maybe I can find an easy patch there).


What I have described in the README is actually a way how to hotfix the issue. 
Maybe that could be phrased better, but I do not know how.

dettus:
Hello.

My apologies... I was not sure if your message meant that you need me to change something.

So, just so that I have asked this question: Do you need me to do anything before you move on with this pull request?

I would REALLY like to see your project running "out of the box" on my favorite operating system. :D

trabucayre:
Hi,
In fact these informations must be moved in specific section of `doc/guide/install.rst`.
Thanks.

dettus:
Understood!
However, after browsing through the doc/guide directory, I found the doc/guid/troubleshooting.rst a more suitable location. 

May I trouble you to have a quick look if this meets your standards?

(P.S.: I am also not familiar with the .rst format.... Sorry.... )

trabucayre:
Good question.
In my mind troubleshooting is more to provides solutions for unusual situations and install is steps to be done to be able to uses openFPGALoader on a specific OS.

dettus:
Oh! I see!

In that case: "troubleshooting" is correct.  :)

I have two boards with a GateMate chip that I wish to access with the openFPGALoader: 
- One from Olimex
- One from CologneChip. 

With Olimex it just works, I think I wouldn't have noticed any issues if I just had that one board.

But for the other one, I have to boot with the reconfigured Kernel.
So, the "Installation" is okay, but the other one is giving me "Trouble". 

Do you agree?

trabucayre:
Applied. Thanks @dettus !

PR #652, opened 2026-04-24
gowin: fix checksum header key typo

viktor-pd:
Header key for checksum parsed from gowin *.fs file is accessed incorrectly (mistyped in src/gowin.cpp). 
Should be CheckSum as created in src/fsparser.cpp:96.


trabucayre:
Applied.
Thanks @viktor-pd !

Issue #646, opened 2026-04-10
Troublesome "passive" mode check in Efinix flash


harbaum:
Is this test really needed?

https://github.com/trabucayre/openFPGALoader/blob/b9486b099e77a75d484d81ca421a292795395f6f/src/efinix.cpp#L221-L225

This causes an exception when trying to flash a bitstream in .bit format as that evaluates to being passive. Flashing .hex files however works as these are identified to be active (or not passive to be precise). But it fails even worse when trying to flash raw files (which I do to load retro ROMs into unused flash areas) as then no mode header value exists at all which in turn makes [getHeaderVal()](https://github.com/trabucayre/openFPGALoader/blob/b9486b099e77a75d484d81ca421a292795395f6f/src/configBitstreamParser.cpp#L97) throw an exception.

Edit: Removing this test fixes all my issues. I can even flash the .bit file, and it's successfully and actively loaded and run by the FPGA. Maybe this should only trigger a warning as there doesn't seem to be any harm when flashing a .bit file.

harbaum:
Fun fact: I decoded both, the .hex and .bit (can easily be done with `xxd -r -ps <file>`) and the only difference is the Mode header line. I wonder [what's the idea behind that](https://forum.efinix.net/t/difference-between-hex-and-bit-file/3878) ...

harbaum:
This test reliably triggers an exception for any raw file flash. Flashing raw files therefore doesn't work at all for Efinix.

harbaum:
Fixed

PR #655, opened 2026-04-27
Efinix: Check for presence of mode header value before checking its value

harbaum:
I must admit that I am not fully aware how raw flashing for efinix is supposed to be handled. But this patch allows flashing of raw files which naturally don't have a "mode" header value at all. This e.g. allows me to flash the Atari ST TOS operating system like so:

` ./openFPGALoader -b trion_t20_bga256 --external-flash -o 0x100000 ./tos104de.img `


harbaum:
This is supposed to fix https://github.com/trabucayre/openFPGALoader/issues/646

trabucayre:
Applied.
Thanks @harbaum !

PR #658, opened 2026-04-30
xilinx: fix to Xilinx::program so a reset is not performed in SPI_MODE when _skip_reset is set

hgomersall:
When writing to the FPGA flash with `--skip-reset` configured on the command line, it was still resetting on completion of the write.

This patch applies a check on `_skip_reset` akin to the check in `Xilinx::post_flash_access` to the `Device::SPI_MODE` path in `Xilinx::program()`.

Testing shows the reset no longer happens and that resets happen as expected in a normal write to memory. It's not clear that there _should_ be two paths to reset (both `post_flash_access` and inside `program()`), but I didn't want to wade into behaviour I didn't fully understand (which indeed I may already have done!).

### DIFF
--- a/src/xilinx.cpp
+++ b/src/xilinx.cpp
@@ -659,7 +659,6 @@ void Xilinx::program(unsigned int offset, bool unprotect_flash)
 		/* Check for BPI flash boards */
 		if (_is_bpi_board) {
 			program_bpi(bit, offset);
-			reset();
 		} else {
 			if (_flash_chips & PRIMARY_FLASH) {
 				select_flash_chip(PRIMARY_FLASH);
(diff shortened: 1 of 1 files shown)


trabucayre:
Hi,
Good catch. But in fact the `reset` call you mention, in `program_spi` context, must be simply removed. `program_spi` calls FlashInterface::write. This method call post_flash_access and this one performs reset if allowed.
For `program_bpi` the code must be improved to do the same thing, but in a first time this part of your PR is fine.

hgomersall:
So to be clear, you want me to remove the reset in the `else` branch?

trabucayre:
I have modified your PR to removes reset calls and added post_flash_access in program_bpi method.

trabucayre:
Applied.
Thanks @hgomersall !

PR #661, opened 2026-05-05
Add --version lowercase flag to comply with convention

parker-research:
Extremely minor improvement, but nearly all program support `<program> --version` to get the version. Supporting `--Version` as titlecase "Version" is good, but supporting the far-more-standard `--version` would be nice.


trabucayre:
Applied with a small fix and a `README.md` update.
Thanks @parker-research !

PR #662, opened 2026-05-06
Fix incorrect handling of spaces and newlines in SVF bitstreams and add progressbar

aisuneko:
Added some missing checks regarding newlines and spaces in SVF files (specifically in svf_jtag.cpp) which, when not present, would cause `SVF_jtag::split_str` to yield such characters when parsing certain malformed SVF bitstreams, crashing the program with `SEGFAULT` and failed assertions on `.empty()` in string vectors.

Also implemented a `ProgressBar` for reading SVF bitstreams.

### DIFF
--- a/src/svf_jtag.cpp
+++ b/src/svf_jtag.cpp
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "jtag.hpp"
+#include "progressBar.hpp"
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied.
Thanks @aisuneko !

PR #663, opened 2026-05-12
feat: add QMTech XC7A100T (FGG676) board support

gHashTag:
## Summary

Adds support for the **QMTech XC7A100T core board** (FGG676 package, MT25QL128 SPI flash) and fixes the root cause that prevented `xc7a100tfgg676`-targeted flash programming from working at all.

### The bug

`spiOverJtag/spiOverJtag_xc7a100tfgg676.bit.gz` was a symlink to `spiOverJtag_xc7a100t.bit.gz`, which is built with the `csg324` pinout. csg324 and fgg676 are different physical packages with different SPI flash routing, so on a real FGG676 board the proxy bitstream drove SPI on the wrong pins. Users hit this as a JEDEC ID readback of `FF FF FE` and high-Z MISO instead of the expected `20 BA 18` for MT25QL128.

The root cause is in `spiOverJtag/build.py`: after building the bitstream for the first package in `packages[family][part]` (here, `csg324`), it `ln -s`'d all other package variants to that same `.bit.gz`. That's fine when the SPI pinout is identical across packages of one device size — but for xc7a100t it isn't.

### Changes

- **`spiOverJtag/constr_xc7a_fgg676.xdc`** — switch to the dedicated SPI configuration pins on XC7A*-FGG676 (per UG470/UG475):
  - `CSn` → `C8`, `MOSI` → `B19`, `MISO` → `A18`, `WPn` → `B18`, `HOLDn` → `A19`
  - `CCLK` is driven internally through `STARTUPE2` (unchanged).
- **`spiOverJtag/build.py`** — accept the full `device+package` form (e.g. `xc7a100tfgg676`) and emit a package-specific bitstream. Only generate per-package symlinks when building the bare-device target, and replace pre-existing symlinks so a package-specific build wins.
- **`spiOverJtag/Makefile`** — register `xc7a100tfgg676` as its own target.
- **`spiOverJtag/README.md`** — document package-specific bitstream builds.
- **`spiOverJtag/spiOverJtag_xc7a100tfgg676.bit.gz`** — removed (was a stale symlink). Regenerate locally with Vivado:
  ```bash
  cd spiOverJtag && make spiOverJtag_xc7a100tfgg676.bit.gz
  ```
- **`src/board.hpp`** — new `qmtechArtix7_100T` board entry (`xc7a100tfgg676`, SPI flash).
- **`doc/boards.yml`** — list `qmtechArtix7_100T` with `Memory: OK` / `Flash: OK`.

### Note on the regenerated bitstream

The repo previously shipped `spiOverJtag_xc7a100tfgg676.bit.gz` as a (broken) symlink. This PR removes that file rather than committing a new one, because Vivado is required to regenerate it and the maintainer's build environment is the canonical place for it. Reviewers / users can build it with the command above.

## Test plan

- [ ] `python3 -c 'import ast, pathlib; ast.parse(pathlib.Path("spiOverJtag/build.py").read_text())'` passes (syntax check).
- [ ] `src/board.hpp` compiles in the openFPGALoader build (verified locally with `g++ -c -std=c++11 -I src`).
- [ ] Rebuild proxy bitstream: `cd spiOverJtag && make spiOverJtag_xc7a100tfgg676.bit.gz` — confirms FGG676 pinout is applied (`set_property PACKAGE_PIN C8 ... csn`, etc.) instead of falling back to csg324.
- [ ] On a QMTech XC7A100T FGG676 board with an MT25QL128 flash, `openFPGALoader -b qmtechArtix7_100T --detect` reports JEDEC ID `20 BA 18` and `--write-flash` / `--read-flash` round-trips successfully.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -688,6 +688,13 @@
   Memory: OK
   Flash: NA
 
+- ID: qmtechArtix7_100T
+  Description: QMTech Artix-7 XC7A100T Core Board (FGG676)
+  URL: https://www.aliexpress.com/item/1005003746671141.html
+  FPGA: Artix-7 xc7a100tfgg676
+  Memory: OK
+  Flash: OK
+
 - ID: qmtechCyclone10
   Description: QMTech Cyclone 10 Starter Kit
   URL: http://www.chinaqmtech.com/productinfo/1858435.html
--- a/src/board.hpp
+++ b/src/board.hpp
@@ -225,6 +225,7 @@ static std::map <std::string, target_board_t> board_list = {
 	JTAG_BOARD("pipistrello",     "xc6slx45csg324",       "ft2232",       SPI_FLASH, 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("pynq_z1",         "xc7z020clg400",        "ft2232",       SPI_FLASH, 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("pynq_z2",         "xc7z020clg400",        "ft2232",       SPI_FLASH, 0, 0, CABLE_DEFAULT),
+	JTAG_BOARD("qmtechArtix7_100T", "xc7a100tfgg676",     "",             SPI_FLASH, 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("qmtechCyclone10", "10cl016484",           "",             SPI_FLASH, 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("qmtechCycloneIVGX", "ep4cgx15027",        "",             SPI_FLASH, 0, 0, CABLE_DEFAULT),
 	JTAG_BOARD("qmtechCycloneIV", "ep4ce1523",            "",             SPI_FLASH, 0, 0, CABLE_DEFAULT),


trabucayre:
The board integration LGTM.
It's interesting for the spiOverJtag bitstream: during modification to have only one bitstream per size, I have tried differents combo and I have observed at the end the SPI interface (die point of view) is always correctly connected to pins (package point of view). Maybe in some specifics case, like this one, this is not true.

trabucayre:
The modifications in `spiOverJtag` directory, `boards.yml` and `boards.hpp` make sense. I'm not convince by the CI integration: the bitstream is built and tested by the contributor so this part is not relevant.
Thanks

trabucayre:
Could you open a new PR for the specific case of the QMTech / bitstream?

gHashTag:
Thanks @trabucayre, done. I split the work as you asked.

1. This PR (#663) is now the plain board integration only. I dropped the CI workflow
   (.github/workflows/build-fgg676.yml) entirely and force-pushed the branch. It is now
   2 files, +8/-0:
   - doc/boards.yml (+7): qmtechArtix7_100T entry
   - src/board.hpp (+1): board_list JTAG_BOARD line
   Head is now 7e60007, rebased clean on master.

2. I opened a separate PR for the QMTech / spiOverJtag specific case:
   https://github.com/trabucayre/openFPGALoader/pull/682
   It carries:
   - spiOverJtag/Makefile: add xc7a100tfgg676 target
   - spiOverJtag/build.py: parse the device+package form and stop aliasing a
     package-specific bitstream to other packages via symlink
   - spiOverJtag/constr_xc7a_fgg676.xdc: set BITSTREAM.STARTUP.STARTUPCLK JtagClk
     (camelCase; the all-caps JTAGCLK literal is silently rejected and falls back to Cclk,
     so EOS stays 0 over JTAG) and correct the SPI pins to LVCMOS33 per the QMTech schematic
   - spiOverJtag/README.md: document the package-specific build
   - removal of the stale prebuilt spiOverJtag_xc7a100tfgg676.bit.gz (the old one was built
     without JtagClk)

That second PR is where the SPI-pin / package edge case you mentioned belongs, so we can
discuss it there. No rush on your side, I can iterate asynchronously.

trabucayre:
Applied.
Thanks @gHashTag !

PR #665, opened 2026-05-14
esp_usb_jtag: support multiple boards (--busdev-num, --esp-serial)

r4d10n:
## Summary
- `libusb_open_device_with_vid_pid()` returns the first VID:PID match, so with two ESP32-S3 cables connected, every invocation programmed the same board. Replace it with `libusb_get_device_list()` iteration that honours `cable.bus_addr` / `cable.device_addr` (already exposed via `--busdev-num`) and a new `--esp-serial` flag that matches the device `iSerialNumber` by substring.
- The `-d /dev/ttyACM*` path cannot help: that's the CDC-ACM interface (iface 0), while JTAG is the vendor iface 2 reached via libusb; the two enumeration orderings are independent.
- `--ftdi-serial` semantics are unchanged (FTDI cables only); the new `--esp-serial` is gated to `MODE_ESP`.

## Test plan
- [x] Single-board path still works (`-c esp32s3 --detect` with one board)
- [x] `--busdev-num <bus>:<dev>` selects each of two connected boards
- [x] `--esp-serial <substring>` selects each of two connected boards (verified with the last byte of the MAC-derived serial: `:84` vs `:94`)
- [x] Bogus serial / bus / mismatched-pair invocations all fail with a clear error echoing the supplied selectors
- [x] Cross-filter test (matching serial of board A combined with bus/dev of board B) is rejected — proves AND semantics across filters
- [ ] Tested only on Linux; macOS/Windows libusb behaviour unverified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,9 @@ openFPGALoader -- a program to flash FPGA
       --cable-index arg         probe index (FTDI and cmsisDAP)
       --busdev-num arg          select a probe by it bus and device number
                                 (bus_num:device_addr)
-      --ftdi-serial arg         FTDI chip serial number
+      --usb-serial-num arg      USB iSerial (FTDI chip serial number or ESP32
+                                iSerialNumber substring)
+      --ftdi-serial arg         FTDI chip serial number (Deprecated)
       --ftdi-channel arg        FTDI chip channel number (channels 0-3 map to
                                 A-D)
   -d, --device arg              device to use (/dev/ttyUSBx)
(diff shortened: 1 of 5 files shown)


trabucayre:
Hi @r4d10n
I have updated your PR with small things:
- rebased branch to current master
- `ftdi-serial` is now deprecated in favor of `usb-serial-num`
- updated the README.md

Could you check/verify the code please?
Thanks

r4d10n:
Looks good 👍

trabucayre:
Applied.
Thanks @r4d10n !

PR #670, opened 2026-05-25
Add option to force finding libftdi1 using pkg-config until MSYS2 pac…

cr1901:
…kage is patched.

The `LibFTDI1Config.cmake` module in MSYS2/MINGW-packages hardcodes paths in a way that `Ninja` (_at least_) cannot find the dependency `libusb` when trying to build OFL. The old `pkg-config` method of finding the dep still works fine, however. So add an option to force the old way until a fix finds its way upstream.

(I'm taking a peek at it, but seems more difficult to fix than I thought.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ option(ENABLE_OPTIM       "Enable build with -O3 optimization level"            
 option(BUILD_STATIC       "Whether or not to build with static libraries"        OFF)
 option(USE_PKGCONFIG      "Use pkgconfig to find libraries"                      ON)
 option(LINK_CMAKE_THREADS "Use CMake find_package to link the threading library" OFF)
+option(LINK_CMAKE_LIBFTDI1 "Use CMake find_package to link the libftdi1 library" ON)
 option(WINDOWS_CROSSCOMPILE "Enable cross-compile on windows"                    OFF)
 option(WINDOWS_STATIC_ZLIB "Link zlib statically for Windows builds"             ON)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -825,7 +826,7 @@ if (LINK_CMAKE_THREADS)
 	target_link_libraries(openFPGALoader Threads::Threads)
 endif()
 
-if (USE_LIBFTDI)
+if (USE_LIBFTDI AND LINK_CMAKE_LIBFTDI1)
```
cr1901, line 829: Hmmm, maybe this isn't the proper fix. When configuring with `pkg-config`, I get:

```
CMake Error at CMakeLists.txt:832 (math):
  math cannot parse the expression: " * 100 + ": syntax error, unexpected
  exp_TIMES (2).
```

But probably better to fix it for `pkg-config` than to gate it out.

trabucayre:
Thanks to point this issue.
In fact I have removed `find_package`: this approach is in many cases error prone.
See https://github.com/trabucayre/openFPGALoader/commit/c6ca4e571b0bb96a310b9ea8a7da38bedc29a882
Could you test with an up to date git to see if this fix your issue?

cr1901:
@trabucayre Your commit works fine. Closing this PR. Maybe in the future, `find_package` can be reintroduced... it seems more integrated w/ `cmake`, so in principle plays nicer with Windows (which aside from MSYS2 doesn't exactly have well-defined library locations). But that can be handled another day.

PR #671, opened 2026-05-26
Fix DNA read on Zynq devices

MYJ695:
Do not apply the Zynq non-volatile memory restriction when only reading FPGA DNA. Allow Zynq devices to use the existing 7-series DNA read path.

Tested on xc7z010 over XVC with --read-dna.


trabucayre:
Applied.
Thanks @MYJ695 !

PR #674, opened 2026-06-05
xilinx: add VU19P (xcvu19p) IR code mapping and dispatch

martinchua:
The xcvu19p is a 4-SLR SSI device whose JTAG instruction register codes
differ from the standard VirtexUltraScale+ mapping. Add a dedicated
`virtexusp_vu19p` IR code table with the correct opcodes (derived from
the xcvu19p BSDL, where SLR1 is the master SLR), and dispatch to it
when the detected model is `xcvu19p`.

Tested on a xcvu19p device.

### DIFF
--- a/src/xilinx.cpp
+++ b/src/xilinx.cpp
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 /*
  * Copyright (C) 2019 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
- * Copyright © 2026 Apple Inc.
+ * Copyright (C) 2026 Apple Inc.
  */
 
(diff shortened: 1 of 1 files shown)


src/xilinx.cpp
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 /*
  * Copyright (C) 2019 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
+ * Copyright © 2026 Apple Inc.
```
trabucayre: Really Apple?
Could you uses (C) to avoid non ascii char?

martinchua: Thanks, changed!

trabucayre:
Applied.
Thanks @martinchua !

PR #679, opened 2026-06-11
Add Passive Serial mode for USB Blaster (Altera/Intel)

gzalo:
I'm reverse engineering one of these boards, https://github.com/gzalo/ncomputing-l130 and it doesn't have the JTAG pins connected to the FPGA, just the passive serial programming ones. Quartus supports using the usb blaster to program it, but apparently it's broken in latest windows versions. This PR adds experimental support for that mode in openFPGALoader.

Validated that it works using that Cyclone I FPGA (ep1c6), let me know if there are any integration concerns!


trabucayre:
Applied.
Thanks @gzalo

PR #681, opened 2026-06-16
spiFlashdb: add ISSI IS25WP256 and Quad Enable for IS25LP/WP256

enjoy-digital:
The IS25WP256 (0x9d7019, 1.8V variant of the IS25LP256) was missing from the database, and neither 256Mb part had its Quad Enable bit described (status register bit 6 per the IS25LP(WP)256D datasheet), making --enable-quad fail on these parts. Required to boot Xilinx bitstreams generated with SPI_BUSWIDTH 4.


trabucayre:
Applied.
Thanks @enjoy-digital !

PR #682, opened 2026-06-18
fix(spiOverJtag): package-specific FGG676 bitstream build for QMTech XC7A100T

gHashTag:
## What

Package-specific spiOverJtag bitstream build for the QMTech XC7A100T (FGG676) core board.
Split out of #663 at the maintainer's request (board integration stays in #663; this PR is
the QMTech / bitstream-specific case).

## Changes

- `spiOverJtag/Makefile`: add the `xc7a100tfgg676` target.
- `spiOverJtag/build.py`: accept the device+package form (e.g. `xc7a100tfgg676`) in addition
  to the bare device, and stop aliasing a package-specific bitstream to other packages via
  symlink (the original symlink logic overwrote a package-specific build).
- `spiOverJtag/constr_xc7a_fgg676.xdc`:
  - `set_property BITSTREAM.STARTUP.STARTUPCLK JtagClk` (camelCase). The all-caps `JTAGCLK`
    literal is silently rejected by Vivado and falls back to Cclk; over JTAG that leaves the
    startup FSM without clocks so EOS stays 0 even though CRC/ID are fine. Verified the prior
    build had COR0 with STARTUPCLK=00 (Cclk).
  - correct the SPI dedicated-config pins to LVCMOS33 per the QMTech XC7A75T/100T/200T core
    board schematic.
- `spiOverJtag/README.md`: document the package-specific bitstream build.
- remove the stale prebuilt `spiOverJtag_xc7a100tfgg676.bit.gz` (it was built without the
  JtagClk fix; it should be rebuilt from the corrected constraints).

## Notes

I could not run a full Vivado build in this environment, so the rebuilt `.bit.gz` is not
included here; the constraints + build.py changes are what produce a correct package-specific
bitstream. Happy to discuss the SPI-pin / package edge case here.


spiOverJtag/constr_xc7a_fgg676.xdc
```diff
@@ -3,11 +3,27 @@ set_property CONFIG_VOLTAGE 3.3 [current_design]
...
+# Table 6-3 lists the valid Vivado literal as `JtagClk` (camelCase) —
+# all-caps `JTAGCLK` is silently rejected and Vivado falls back to Cclk.
+set_property BITSTREAM.STARTUP.STARTUPCLK JtagClk [current_design]
+
+# XC7A75T/100T/200T-FGG676 SPI flash pins on QMTech XC7A100T-2FGG676I core board.
+# Verified against QMTECH_XC7A75T_100T_200T-CORE-BOARD-V01-20210109.pdf schematic
+# (https://github.com/ChinaQMTECH/QMTECH_XC7A75T-100T-200T_Core_Board).
+# These are Bank 14 dual-purpose user IO pins (D00..D03, FCS_B) wired to the
+# on-board N25Q064A SPI flash (JEDEC 0x20BA17). CCLK is driven internally via
+# the STARTUPE2 primitive instantiated in xilinx_spiOverJtag.v.
```
trabucayre: It's maybe too specific: These pins are used for SPI flash in all model with this package. Not only for QMTech boards

gHashTag: You're right, thanks. The SPI-flash balls are fixed by the FGG676 package, not by the
QMTech board, so scoping the comment to QMTech was misleading. I reworded it to say these
are package-level pins for any XC7A*T-FGG676 part and kept the QMTech schematic only as the
cross-check reference, not as the scope. No pin values changed. Pushed as 2b024ce.

trabucayre:
Applied.
Thanks @gHashTag !

PR #683, opened 2026-06-19
xvc_client: block until full reply received, fixes errors on a slow network

pftbest:
ll_write() can request more than 1.5K of data, but recv() call will return right away as soon as one packet is received.
This causes large transfers to break as not enough data is received.

Fixed by adding rx_exact flag and enabled it for ll_write()

Tested with Zynq7010 on a slow WiFi. Before the patch all uploads were silently failing with no error in the log.


trabucayre:
Applied.
Thanks @pftbest !

PR #685, opened 2026-06-26
Xcau7p support

poleguy-shure:
Adding support for three SPI Flash chips not supported by vivado for indirect programming.
Added support for xcau7p-sbvc484 (using vivado 2025.2).
Improved a couple failure cases that didn't return error codes that tripped me up during testing.

This has been tested on hardware to prove read/write works on one of these three chips. I will continue to test other functionality after the other chips are soldered on to a test board.

### DIFF
--- a/spiOverJtag/build.py
+++ b/spiOverJtag/build.py
@@ -166,6 +166,7 @@ if tool in ["ise", "vivado"]:
         "xcku3p-ffva676"   : "xcku3p_ffva676",
         "xcku3p-ffvb676"   : "xcku3p_ffvb676",
         "xcku5p-ffvb676"   : "xcku5p_ffvb676",
+        "xcau7p-sbvc484"   : "xcau7p_sbvc484",
         "xcau10p-ffvb676"  : "xcau10p_ffvb676",
         "xcau15p-ffvb676"  : "xcau15p_ffvb676",
     }.get(part, pkg_name)
@@ -233,7 +234,7 @@ if tool in ["ise", "vivado"]:
                     "description" : "secondary flash",
                     "default"     : 1,
                 }
-            elif part == "xcau10p-ffvb676":
+            elif part in ["xcau10p-ffvb676", "xcau7p-sbvc484"]:
                 tool_options = {"part": part + "-1-e"}
             elif part == "xcau15p-ffvb676":
                 tool_options = {"part": part + "-2-e"}
--- a/spiOverJtag/xilinx_spiOverJtag.v
+++ b/spiOverJtag/xilinx_spiOverJtag.v
@@ -32,6 +32,13 @@ module spiOverJtag
 	wire capture, drck, sel, update, shift;
 	wire tdi, tdo;
 
+`ifdef xilinxultrascale
+	/* For ultrascale these signals are internal wires (not module ports);
+	   declare them here so they are visible to spiOverJtag_core_prim below. */
+	wire csn;
+	wire sdi_dq0, sdo_dq1, wpn_dq2, hldn_dq3;
+`endif
+
 `ifndef spartan3e
 `ifndef virtex6
 	/* Version Interface. */
@@ -75,7 +82,6 @@ module spiOverJtag
 `else // !spartan6 && !spartan3e
 
 `ifdef xilinxultrascale
-	assign sck = drck;
 	wire [3:0] di;
 	assign sdo_dq1 = di[1];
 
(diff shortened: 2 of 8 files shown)


spiOverJtag/build.py
```diff
@@ -233,6 +234,8 @@
                     "description" : "secondary flash",
                     "default"     : 1,
                 }
+            elif part == "xcau7p-sbvc484":
+                tool_options = {"part": part + "-1-e"}
             elif part == "xcau10p-ffvb676":
                 tool_options = {"part": part + "-1-e"}
```
trabucayre: Maybe:
```python
elif part in ["xcau7p-sbvc484", "xcau10p-ffvb676"]:
    tool_options = {"part": part + "-1-e"}
```
Is a better/simplier way

spiOverJtag/xilinx_spiOverJtag.v
```diff
@@ -75,7 +75,8 @@ module spiOverJtag
 `else // !spartan6 && !spartan3e
 
 `ifdef xilinxultrascale
-	assign sck = drck;
+	wire csn;
+	wire sdi_dq0, sdo_dq1, wpn_dq2, hldn_dq3;
```
trabucayre: This signals must be defined before line 41 (ie before spiOverJtag_core instanciation) because these are used by this sub module

trabucayre:
Looks good to me. Just few remarks.
Thanks @poleguy

spiOverJtag/build.py
poleguy-shure: I think the purpose of this file is not very clear to me. (not well commented?). Nor is it clear to me exactly what the purpose of this tool_options assignment is.

I suspect much of this file should really be a data structure rather than a billion lines of conditional code. This section seems to map the dumbed-down part name and package info into a full name as required by vivado. (?)

If that's the case, I don't think combining these lines makes the intent clearer. There is no connection between the two parts you're suggesting putting together, so I think they should remain independent to make it clear this is just a simple name mapping?

spiOverJtag/xilinx_spiOverJtag.v
poleguy-shure: This file is hard for me to follow with all the ifdef options. I wonder if breaking it into different files for the different cases would be easier to maintain?

I don't think technically verilog cares which order the declarations are in, so it doesn't technically have to move. But from a style point of view it is fine to have them up top, so I'll move them.

One weird thing is that csn is conditionally an output depending on an ifndef at line 5. I can't say I fully understand this.

poleguy-shure:
Thanks for managing this project! It is a great feeling to be able to contribute support for SPI Flash chips that Xilinx doesn't bother supporting. 

My company recently introduced a policy to formally allow for contributions to open source projects. I'm working through step 6 of the 8 step process at the moment. I hope to give some feedback to my management about how this went.

I also have code to support the flash on the Alchitry Pt board. I compiled it for the raspberry Pi, and it works great. I'll try to contribute that as well in the future.

poleguy:
Can this be merged as is? Or do you feel there is something more I should address first?

spiOverJtag/build.py
trabucayre: I am agree: I'm not good for documentation.
`build.py` uses `edalize` as backend to deal with vendor tools. `part` is the name of the device with the speed grade (here -1-e). Naming depends on devices and the slowest speedgrade is better to match all variant to a device.
With a list the meaning is "for these devices the speed grade is xxx" instead of "for this device the speedgrade is xxx, for this one xxx too, for this another one yyy".
This why I prefer a list instead of a too long list of if/else/elif

spiOverJtag/xilinx_spiOverJtag.v
trabucayre: I'm agree this file is a bit complex. Most of the devices have a logic similar and to avoid to have dupplicate code with the needs to have incoherency only one verilog is used for all devices.
But I think I have to see how to simplify to avoid has much as possible inner ifdef

trabucayre:
Thanks for your feedback. I think I will updates this PR by myself to fix nitpick before merging.
Happy to see companies have a policy to allows contributions in open source projects!
Feel free to contributes remaining flash.

trabucayre:
Could you recheck update I have performed in your branch?
Thanks!

trabucayre:
Applied.
Thanks @poleguy-shure !

poleguy:
I plan to take another look at this on my hardware this week. I'll let you know if everything checks out.

PR #676, opened 2026-06-08
Add support for alibaba cloud xcku3p card

boondocklabs:
Created spiOverJtag for the xcku3p-ffvb676
Added `alibaba_u3p` board

PR #686, opened 2026-07-01
xilinx: add BPI parallel NOR support to dumpFlash()

haomjc:
## Summary
- BPI (parallel CFI NOR) flash support was added in #676 for `program_bpi`/write, but `Xilinx::dumpFlash()` only handles the XC95/XCF and SPI (`PRIMARY_FLASH`/`SECONDARY_FLASH`) paths.
- On boards where `_is_bpi_board` is set, `--dump-flash` currently falls through to the SPI `FlashInterface`, silently producing garbage instead of reading the BPI NOR.
- This adds a BPI branch mirroring the existing write path: load the BPI JTAG bridge if needed, `detect()` the flash, `read()` the requested region, and write it to the output file.

## Test plan
Tested on real hardware: a Kintex-7 XC7K480T board with a Micron MT28GU512AAA1EGC BPI NOR, via a Digilent HS2 cable (`-c digilent_hs2 -b <board>`).

- [x] Wrote a block of random data to a scratch flash region with `--write-flash -o <offset>`.
- [x] Read the same region back with the patched `--dump-flash -o <offset> --file-size <len>`.
- [x] Confirmed the written and read-back files are byte-identical (`cmp`).
- [x] `dumpFlash()` correctly calls `load_bpi_bridge()` when `_bpi_flash` isn't already loaded, and `post_flash_access()` to restore JTAG state afterward, matching the pattern used elsewhere in the BPI write path.

Note: BPI reads are one word per USB round-trip (no burst read in `BPIFlash::read()`), so large dumps are slow — this PR doesn't change that, it just makes `--dump-flash` functional for BPI boards.

### DIFF
--- a/src/xilinx.cpp
+++ b/src/xilinx.cpp
@@ -841,6 +841,35 @@ void Xilinx::program_bpi(ConfigBitstreamParser *bit, unsigned int offset)
 	printInfo("BPI flash programming complete");
 }
 
+bool Xilinx::dumpFlash_bpi(uint32_t base_addr, uint32_t len)
+{
+	if (!_bpi_flash) {
+		if (!load_bpi_bridge())
+			return false;
+	}
+	if (!_bpi_flash->detect()) {
+		printError("BPI flash detection failed");
+		return false;
+	}
+	if (len == 0)
+		len = _bpi_flash->capacity();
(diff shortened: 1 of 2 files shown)


trabucayre:
Applied with a small modification.
Thanks @haomjc !

PR #691, opened 2026-07-12
fix SST26VF064B erase

moffel:
The SST26VF064B is just like the fix SST26VF032B in that it has a block erase command but no uniform block size across the entire device. This results in only the first 8 KiB beeing erased while the first 64 KiB was intended.


trabucayre:
Applied.
Thanks @moffel !

PR #688, opened 2026-07-04
boards: add ALINX AXKU095 (XCKU095-FFVA1156)

davekeeshan:
Add board preset for the Alinx AXKU095 Kintex UltraScale dev board, using the Digilent HS2 cable (FT232H). Tested JTAG detect and SRAM programming with IDCODE 0x23844093.

https://www.en.alinx.com/Product/FPGA-Development-Boards/Kintex-UltraScale/AXKU095.html

there are two other boards of exactly the same form factor, I imagine it is relativity easy to include them, however i do not have access to them so i can not verify them, I have only physically verified it on the AXKU095 

https://www.en.alinx.com/Product/FPGA-Development-Boards/Kintex-UltraScale/AXKU062.html
XCKU060-2FFVA1156I

https://www.en.alinx.com/Product/FPGA-Development-Boards/Kintex-UltraScale/AXKU042.html
XCKU040-2FFVA1156I

PR #693, opened 2026-07-14
boards: add ALINX AXKU042 and AXKU062 dev boards

davekeeshan:
Thanks for merging the AXKU095 board in #688,  here is a PR with the 2 other boards, but as I indicated previously i don't have them, so I am not able to verify them. However based on the form factor I am pretty sure they will work.  However I will leave it up to you, if this path is considered verboten just refused and squash the PR.

### DIFF
--- a/doc/boards.yml
+++ b/doc/boards.yml
@@ -34,6 +34,20 @@
   Memory: OK
   Flash: OK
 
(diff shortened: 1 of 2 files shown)


trabucayre:
Applied.
Thanks @davekeeshan !

PR #696, opened 2026-07-24
Doc: fix exclusive issue in example commands

mer0m:


trabucayre:
Applied.
Thanks @mer0m !

PR #697, opened 2026-07-30
Add Altera Cyclone II device IDs.

lhartmann:
As listed on Table 3–3. 32-Bit Cyclone II Device IDCODE, from Cyclone II Device Handbook, Volume 1.

This enables upload of SVF files to Altera/Terasic DE2 educational boards (legacy, but still useful).

Tested EP2C35 on the DE2. Others untested.


trabucayre:
Applied.
Thanks @lhartmann !

Issue #704, opened 2026-08-12
Native win32 support (win32 headers, msvc/clang)


fredemmott:
- [x] Do not depend on GCC extensions
- [x] Do not unconditionally add GCC flags
- [x] Do not depend on mingw/cygwin libraries or userspace
- [x] Build with VS 2022
- [x] Build with VS 2026 (generally stricted than VS 2022)
- [x] Dependencies
- [ ] `D:\openFPGALoader\src\ch347jtag.cpp(58): warning C4477: 'fprintf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type '__int64'`

I've locally got a locally patched v1.1.1 building with VS2022 with support for the gwu2x only; I've filed this issue primarily as a place for me to attach pull requests to.

fredemmott:
WIP here, will send separate PRs once more polished/complete: https://github.com/fredemmott/openFPGALoader/tree/msvc

fredemmott:
Mostly done on splitting my branch into reviewable PRs, but taking a bit of a break.

Current branch state is a VS 2022 build with all cables included. Tested with GWA25X (ModRetro Chromatic). Not tested VS 2026 yet.

fredemmott:
Okay, I *think* that's everything done.

Once the PRs are in, would y'all like a github actions config for this?

PR #705, opened 2026-08-12
MSVC: Set C++ standard to 20

fredemmott:
Designated initializers are used, which is a C++20 feature (or GCC extension)

refs #704

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ endif()
 ####################################################################################################
 
 ## specify the C++ standard
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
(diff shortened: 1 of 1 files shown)


fredemmott:
A potential followup to this would be using `std::format` to avoid issues like this:

```
D:\openFPGALoader\src\ch347jtag.cpp(58): warning C4477: 'fprintf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type '__int64'
D:\openFPGALoader\src\ch347jtag.cpp(58): note: consider using '%lld' in the format string
D:\openFPGALoader\src\ch347jtag.cpp(58): note: consider using '%Id' in the format string
D:\openFPGALoader\src\ch347jtag.cpp(58): note: consider using '%I64d' in the format string
```

PR #709, opened 2026-08-12
Use `std::filesystem` for win32 absolute path conversions

fredemmott:
Avoids dependency on cygpath.

Part of C++17, so *does not* depend on #705 

refs #704


fredemmott:
It *might* be best to do this on all platforms, not just win32?

PR #717, opened 2026-08-12
MSVC: Replace C99 dynamic-sized arrays with std::vector

fredemmott:
While they are in C99, they are not technically standard C++, and MSVC does not aim to support C99.

refs #704


fredemmott:
Created as a draft as there's lots of 'boring' changes here, and I'd like to take a closer look for typos etc before asking y'all to spend your time reviewing it.

PR #718, opened 2026-08-12
Add toolchain file to optimize MSVC-based release builds

fredemmott:
- use 'hybrid CRT' to keep it dependency-free on Win10+ without the full size of static linking
- add default MSVC release optimizations to RelWithDebInfo

refs #704

---

Testing:

```
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=D:/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=D:/openfpgaloader/cmake/Toolchain-x86_64-w64-msvc-redist.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-static

cmake --build . --parallel --config RelWithDebInfo
```

On my v1.1.1 branch, this gets me a 2.5MB exe that only depends on:
- `WS2_32.dll` (winsock2, specifically `ntohs()`
- `KERNEL32.dll`
- `ucrtbase.dll` (~ libc)

... and also gives me ~ 20MB of debug symbols, separate to the exe:

image

I then used this to re-flash a ModRetro Chromatic with customized firmware:

image

### DIFF
--- a/cmake/Toolchain-x86_64-w64-msvc-redist.cmake
+++ b/cmake/Toolchain-x86_64-w64-msvc-redist.cmake
@@ -1,0 +1,30 @@
+# Optimize for small, fast, and dependency-free exes
+
+if (VCPKG_TARGET_TRIPLET AND NOT VCPKG_TARGET_TRIPLET MATCHES "-static$")
+  message(FATAL_ERROR "${CMAKE_CURRENT_LIST_FILE} requires a -static vcpkg triplet")
+endif()
+
+# We're going to use the 'Hybrid CRT' approach, which is the combination of the
+# UCRT and the static C++ Runtime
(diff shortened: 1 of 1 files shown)


fredemmott:
Just for comparison, here's the current mingw build (~ 4MB):

image

... and the mingw ucrt build (~ 4MB):

image

By comparison, this toolchain + MSVC allows 'just download the exe', more suitable for integration with other programs (e.g. FlashGBX) - and it's ~ half the size of the mingw binaries that depend on libftdi/libusb

PR #724, opened 2026-08-26
Update lattice.rst

kaborja:
Correct RST and DONE pin numbers on the FTDI


trabucayre:
Applied.
Thanks @kaborja !

PR #726, opened 2026-09-04
Added support for xcvu9p-flgb2104 with 2 flash configuration in quadspi

Alenlu:
Added support for mentioned fpga, also added a device xc2c256 in parts.
Changes in s25fl512s to support quadspi bit, like described in https://www.infineon.com/assets/row/public/documents/10/49/infineon-s25fl512s-military-512-mbit-64-mbyte-3-datasheet-en.pdf.
Also added a bigger timeout in bulk_erase, to handle bigger flash devices.

All was tested in multi chain configuration.


trabucayre:
Applied.
Thanks @Alenlu !

PR #730, opened 2026-09-14
Support DLC10

dayo05:
I have Spartan7 with Xilinx Platform Cable USB II(DLC10).

I faced issue that even JTAG DLC10 cable ships base firmware on EEPROM. 
But, current code treats 0x0013 as uninitialized, so it tries to check XPCU_INITIALIZED_PID that leads timeout error.

This PR does
1. Skip firmware check on id == 0x0013
2. Skip FX2 firmware injection on id == 0x0013

I only tested this with MacOS 26.6.2 and not tested on other OS.


trabucayre:
The problem with this PR is all cables uninitialized with PID = 0x13 (like the Waveshare DLC10) will fails because the firmware will not be loaded.
I think is required to have a more clever approach to deal with this specific use case.
Could you provides the lsusb -vv -d 03fd:0013 (or the similar thing for macos)?
It's maybe required to check `idProduct`, `iManufacturer` or something like this to detect if the firmware must be loaded (or not) when the PID is 0x13.

PR #731, opened 2026-09-15
svf_jtag: fix error output

TuxCoder:
this appends a leading 0 to bytes printed,
otherwise output gets missing zeros in the middle.

my c++ std::ostream skills are not the best, but I think this patch should work.


trabucayre:
Applied.
Thanks @TuxCoder !

PR #734, opened 2026-09-20
spiFlashdb: add GigaDevice GD25Q16C (0xc84015)

reinauer:
The Colorlight i5 (ECP5 LFE5U-25) carries a **GigaDevice GD25Q16C**, a 16Mbit (2MB) SPI NOR reporting JEDEC ID `c8 40 15`, which is not yet in `spiFlashdb.hpp`:

```
Jedec ID          : c8
memory type       : 40
memory capacity   : 15
```

With no match, `_flash_model` stays `NULL` and `display_status_reg()` falls back to generic status register decoding, so block protection handling is not datasheet accurate.

### The entry

* 32 sectors of 64KB (2MB total)
* 64KB sector erase and 4KB subsector erase
* `BP[3:0]` at bits 2..5 of the status register, `CMP` at S14 — same as the neighbouring GD25Q32C entry
* Not OTP, no global lock


trabucayre:
Applied.
Thanks @reinauer !

